PR #7, opened 2015-08-15
:bug: Format string conflates dates/minutes

kjgorman:
MM is a two digit month representation (e.g. August -> "08"), whereas mm
is a two digit minute representation.


hassy:
Cheers @kjgorman :+1:

PR #13, opened 2015-10-12
Quiet option

fabiosantoscode:
This came up while @xicombd was writing something using minigun.

It is unshibelike to not say anything, but this PR adds the `--quiet`/`-q` option to the command line, which can also be passed through javascript.

Although it might not be so useful on the command line, scripts using minigun as a library will want to stop it from writing to stdout.

### DIFF
--- a/bin/minigun
+++ b/bin/minigun
@@ -16,6 +16,7 @@ program
           'perform "insecure" TLS connections, for testing with' +
           'self-signed certificates')
   .option('-e, --environment <name>', 'Specify the environment to be used')
+  .option('-q, --quiet', 'Do not print anything to stdout')
   .action(commands.run);
 
 program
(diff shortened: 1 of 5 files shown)


xicombd:
:+1:

hassy:
doge

Thanks! :+1: This is a needed option.

Changes that need to be made before merging:
- there is no need to pass quiet to core (it should not output anything to stdout/stderr)
- `dino` does not need a quiet option :)

xicombd:
Hey @hassy! Where are we actually passing the quiet option to core?

fabiosantoscode:
@xicombd was kind enough to PR into my PR

hassy:
Thanks guys! :)

shibe

PR #22, opened 2015-11-20
Recommend wrk instead of ab for http testing.

tomgco:
ab is old, single threaded and very basic, wrk allows some basic lua scripting as well and I find it much more stable. 

`ab` also has the issue that on certain OS X installations it is broken.


hassy:
:+1: cheers

Issue #23, opened 2015-11-24
There should be a way to define the target from a cmdline argument or an env variable


Soarez:
Any guidelines for a PR on this?

hassy:
This would be awesome. How about a command line switch to `minigun` that allows the user to overwrite any field in the config file. Something like this to override `config.target`.

```sh
minigun run --config 'target=http://app.local' myscript.json
```

or like this to override a default header:

```sh
minigun run --config 'defaults.headers["x-custom-header"]=1234567890' myscript.json
```

An perhaps cleaner alternative is to use JSONPath to select the path to be modified (since JSONPath is already used for parsing and capturing responses inside scenarios). Something like this:

```sh
minigun run --config '$.target' 'http://app.local' myscript.json # where $ is the config object
```

Thoughts?

Soarez:
[`commander`](https://www.npmjs.com/package/commander) offers no support for using multiple instances of the same command line flag. i.e. it wouldn't be possible to

```bash
minigun run --config '<override target>' --config '<override headers>' my script.yml
```

So unless we go with another cmdline args parser, we will only be able to override one option with this interface.

I only need to override the target, I'm not sure if overriding other things is really all that necessary.

The reason I need to override the target is because multiple machines are running the same API, and I routinely need to bench them and compare them. Right now, it's very hard to script this. If you think this scenario might not be that uncommon, then maybe:

```bash
minion run myscript.yml optional_target
```

or

```bash
minigun run -p target myscript.yml
```

These would override the target on the script and can also make them not necessary i.e. you could define them in either place.

What do you think?

hassy:
Can we make it `--target`? +1 on actually not needing to be able to override everything yet.

PR #24, opened 2015-11-24
Target option

Soarez:
Re #23


hassy:
:+1: cheers

shibe.js

PR #26, opened 2015-11-27
Now errors when no target specified

lukebond:
con tests


hassy:
:+1: cheers

shiba.js

PR #40, opened 2016-01-17
Use rc for default config values

xicombd:
This allows you to specify config default values for the `run` command using a `.minigunrc` file on your home folder or in any other [location supported by `rc`](https://github.com/dominictarr/rc#standards). 

The included `.minigunrc` serves as an example and is used for the test as well.

Let me know what you think :)

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -10,10 +10,13 @@ const csv = require('csv-parse');
 const util = require('util');
 const cli = require('cli');
 const YAML = require('yamljs');
+const defaultOptions = require('rc')('artillery');
 
 module.exports = run;
 
 function run(scriptPath, options) {
+
+  options = _.defaultsDeep(options, defaultOptions);
 
   let logfile =
     'artillery_report_' +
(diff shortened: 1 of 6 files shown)


hassy:
Thank @xicombd! I like the idea of being able to customize settings via a conf file, but for running actual load-tests it doesn't make sense to customize behavior of a test script with an external config file.

So:
1. Having `rc` is useful :+1: 
2. The conf file should not be used to over-ride settings in test scripts - could you remove that code?
3. A good use case is setting the format for log names for example, or enabling the Blessed UI by default. :)

xicombd:
Thanks for the feedback!

> 2. The conf file should not be used to over-ride settings in test scripts - could you remove that code?
How would you want to do that? Minigun only uses the values from the file, if they are not on the test file, and the files have their own `phases ` specified, so only the new test should be affected by it.
Or am I missing something?

lib/commands/run.js
```diff
@@ -59,6 +60,8 @@ function run(scriptPath, options) {
           return callback(new Error(msg2), null);
         }
 
+        script.config = _.defaultsDeep(config, script.config);
```
hassy: @xicombd this line here is what I meant, it allows the user to implicitly override settings in the actual script file with values from .minigunrc

PR #66, opened 2016-03-16
Allow output-only for html reports. Fixes #65

### LINKED ISSUE #65: Write html report to file, don't open. For CI
Something that I want to do is just make the html file available for downloading from my CI environment, and not try to open it in a browser.

Opening the browser means my build never ends, and fails.

I propose passing -o/--output to the report command, as with the run command, and if specified, output to  a file and don't try to open the report.

antony:


hassy:
:+1: nice

PR #78, opened 2016-04-13
Invalid file message is now adjusted to the file format.

Nepoxx:
Invalid file format message was always complaining about JSON even if the file was YAML.


hassy:
👍  nice

Issue #29, opened 2015-12-18
minigun dino: different dino’s in the same day.


tomgco:
This is conflicting with the `--help` information: 

```
    dino [options]          Show dinosaur of the day
```

```
$ minigun dino
 __________
< minigun! >
 ----------
          \
           \
            __
           / _)
    .-^^^-/ /
 __/       /
<__.|_|-|_|

$ minigun dino
 __________
< minigun! >
 ----------
          \
           \
               __
              / _)
     _/\/\/\_/ /
   _|         /
 _|  (  | (  |
/__.-'|_|--|_|

```

hassy:
Looks like a serious issue in the day of the week -> dino hashing function. :cold_sweat:

hassy:
We could do with more dinos too. :dragon:

tomgco:
> We could do with more dinos too.

Excited for 1.4 now :dancer:

PR #89, opened 2016-05-13
(#29) - One Dino Per Day

erikerikson:
Use the localized year, month, and day modulo the length of the dino art to
select the proper art for the day.

Add brontosaurus and tyrannosaurus.
Potential issue: misalignment to speech bubble.
Hoping no one was feeling strongly about simple dino art.

### DIFF
--- a/lib/commands/dino.js
+++ b/lib/commands/dino.js
@@ -4,8 +4,6 @@
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


hassy:
That is some quaity lo-fi dino art! 👏

Issue #72, opened 2016-03-22
CSVs don't appear to work


antony:
I realise there is a test for this that seems to pass, but given my CSV:

```
abcde,Mita,20554,Novick,Septfontaines,1111,asnyder0@dagondesign.com,Tinactin,Andrea Snyder,asnyder0@wp.com
1243fsdfs,Browsezoom,385,Logan,Malkā,4444,levans1@51.la,SORE MUSCLE AND JOINT PAIN RELIEF,Lawrence Evans,levans1@desdev.cn
gdfgdfg,Flashset,9451,Sage,Amarillo,79165,ebell2@desdev.cn,Morphine Sulfate,Eugene Bell,ebell2@blog.com
```

and my config block:

```
    "payload": {
      "path": "./data2.csv",
      "fields": ["registration","name","building","street","city","postcode","email","branch","fullName","fullEmail"]
    }
```

My request is:

```
  {
          "loop": [
            {
              "post": {
                "url": "/v1/xxyyzz/abcde",
                "json": {
                	"registration": "{{ registration }}",
                	"name": "{{ name }}",
                	"contact": {
                		"address": {
                			"building": "{{ building }}",
                			"street": "{{ street }}",
                			"line1": "1th 4th",
                			"city": "{{ city }}",
                			"postCode": "1234",
                			"country": "DK"
                		},
                		"phoneNumber": "12345678",
                		"email": "{{ email }}"
                	}
                },
                "capture": {
                  "json": "$._id",
                  "as": "someId"
                }
              }
            }
          ],
          "count": 1
        }
```

I'm having no luck at all - unfortunately I can't see a way to get good debug information about the requests i'm sending (http:request would be nice) but my API is telling me that 'registration' is blank each time, and if I hardcode it, the next entry is blank, and so on.

hassy:
Thanks for the bug report! This is a known bug, for now the workaround is to wrap the payload definition in an array like this:

```
    "payload": [{
      "path": "./data2.csv",
      "fields": ["registration","name","building","street","city","postcode","email","branch","fullName","fullEmail"]
    }]
```

hassy:
This is the offending line btw if you want to take a stab at it:

https://github.com/shoreditch-ops/artillery/blob/master/lib/commands/run.js#L105-L107

`options.payload` is the path to the payload file passed in via the `-p` flag. The code needs to take into account that `options.payload` might be undefined but the payload definition may have the `"path"` attribute set.

[...]

PR #100, opened 2016-05-20
Issue 72 csv

enderv:
Fix for #72 . Adjusted tests to cover just fields being specified in json payload with path in command line, and added another for both in the json payload.

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -109,13 +109,12 @@ function run(scriptPath, options) {
                  (_.isObject(context.script.config.payload) ||
                  options.payload)) {
         //use config if path set else use command line
-        let csvdata;        
-        if (context.script.config.payload.path){
-           csvdata = fs.readFileSync(context.script.config.payload.path, 'utf-8');
-        } else{
-           csvdata = fs.readFileSync(options.payload, 'utf-8');
+        let csvdata;     
+        if (context.script.config.payload.path) {
+          csvdata = fs.readFileSync(context.script.config.payload.path, 'utf-8');
+        } else {
+          svdata = fs.readFileSync(options.payload, 'utf-8');
         }
-        
         csv(csvdata, function(err, payload) {
 
           if (err) {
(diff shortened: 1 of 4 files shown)


lib/commands/run.js
```diff
@@ -106,9 +106,16 @@ function run(scriptPath, options) {
...
       } else if (context.script.config.payload &&
-                 _.isObject(context.script.config.payload) &&
-                 options.payload) {
-        let csvdata = fs.readFileSync(options.payload, 'utf-8');
+                 (_.isObject(context.script.config.payload) ||
+                 options.payload)) {
+        //use config if path set else use command line
+        let csvdata;        
+        if (context.script.config.payload.path){
+           csvdata = fs.readFileSync(context.script.config.payload.path, 'utf-8');
```
hassy: The path to payload file is meant to be relative to the location of the test script, whereas this will read it relative to the current working directory of the process.

lib/commands/run.js
```diff
@@ -106,9 +106,16 @@ function run(scriptPath, options) {
...
+        } else{
```
hassy: Formatting (missing spaces)

lib/commands/run.js
```diff
@@ -106,9 +106,16 @@ function run(scriptPath, options) {
             return callback(null, context);
           });
       } else if (context.script.config.payload &&
-                 _.isObject(context.script.config.payload) &&
-                 options.payload) {
-        let csvdata = fs.readFileSync(options.payload, 'utf-8');
+                 (_.isObject(context.script.config.payload) ||
+                 options.payload)) {
```
hassy: Don't think we need this check (as we're using `options.payload` if it's available)

hassy:
Looks good 👍  Good to merge once the inline comments have been addressed.

lib/commands/run.js
enderv: saying just remove the `|| options.payload` then? Wasn't sure if it would still enter this block without that bit.

hassy:
Great! Thanks @enderv

PR #172, opened 2016-09-14
Minor fixes

hassy:


lib/commands/run.js
```diff
@@ -34,16 +34,18 @@ function run(scriptPath, options) {
     }
   }
 
-  if (!isDir) {
+  const defaultFormat = '[artillery_report_]YMMDD_HHmmSS[.json]';
+  if (!isDir && options.output) {
```
jenriquejr: 👍 was about to log this as a bug

hassy:
@jenriquejr `v1.5.0-14` is now out with the fix

PR #179, opened 2016-09-22
Expose commands to Node modules

### LINKED ISSUE #173: Programmatic API
I'm trying to develop a artillery runner which can run artillery json for arbitrary modules. I'm currently spawning child processes and executing the script via CLI. But doing all that in a cross platform way is a real pain and buggy.

Would be great to have the artillery commands exposed for consumption via Javascript. Specifically, something like:

```js
var artillery = require('artillery')
artillery.run(...)
artillery.report(...)
...
```

mrchief:
Fixes #173 

@hassy I think this should do the trick. I'm on Windows so couldn't verify with all the make/test commands. I don't anticipate this breaking anything but you never know! :)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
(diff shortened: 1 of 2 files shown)


hassy:
Nice 👍

varatep:
Are there docs for using this or looking for help on that as well?

hassy:
@varatep no docs, PRs welcome

PR #201, opened 2016-11-08
report option -t "Truncate latencies"

servusoft:


hassy:
Thanks for the PR @servusoft! A better solution than an extra command line flag would be to not include every latency measurement in the HTML in the first place (they're not getting used anyway). By the way, this PR and #200 should've been combined as they introduce one logical change.

servusoft:
Hi @hassy, thanks  for reply.
I made the change with browser and could't found, how to combine two changes as one logical change.
The sampled latencies in json may be omited to. its a good idea. 
For my was simpler to truncate it in HTML-report.

hassy:
@servusoft if you send a PR which omits the full latency report from the HTML by default I'll merge that. For now I'm closing this and the other PR. In future to avoid unnecessary work it may be a good idea to open an issue first.

servusoft:
@hassy I try to build the 'omit latency' flag at JSON generator.
Probely would better to have a 'include latency' flag.

I think for more flexibility it should be possible both flags. (at generetor and report)
A latency diargam builds from row data may be compared with build-in diagram.

PR #203, opened 2016-11-10
bugfix numWorkers

servusoft:
`let numWorkers = process.env.ARTILLERY_WORKERS || 1 || os.cpus().length;`
has always the number of workers = 1. 
after exchage ` os.cpus().length || 1;` contains `numWorkers` the number of cores.


hassy:
Thanks @servusoft. This is intentional while multicore support is still an experimental feature.

PR #259, opened 2017-02-17
Report percentage by scenarios created not completed

kjgorman:
Hey Hassy,

When a report is run for a test that produces some number of errors, the percentages shown for different flows start to look a little weird:

image

I think the percentages should be derived from all scenarios created, not just those completed, which should make the percentages add back up to 100%.

What do you think?


hassy:
Nice, thanks @kjgorman!

PR #265, opened 2017-03-01
Add file upload feature

zcei:
After https://github.com/shoreditch-ops/artillery-core/pull/148 landed in master, it's now time to add support to the CLI.

As a reference, the copy-pasted intended usage:

## Examples of Input

Include them in config:

```yaml
config:
  files:
    - "@./path/to/file"
    - "@../path/to/other/file"
```

Reference them from within your scenario

**Single File**

```yaml
post:
  url: "/test"
  body: "@./path/to/file"
```

**Multiple Files - Multipart Form**

```yaml
post:
  url: "/test"
  formData:
    foo: "@./path/to/file"
    bar: "@../path/to/other/file"
```

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -13,6 +13,7 @@ const createRunner = core.runner;
 const debug = require('debug')('artillery:worker');
 
 const path = require('path');
+const fs = require('fs');
 
 process.on('message', onMessage);
 let shuttingDown = false;
@@ -67,6 +68,20 @@ function run(opts) {
     let processorPath = path.resolve(path.dirname(absoluteScriptPath), opts.script.config.processor);
     let processor = require(processorPath);
     opts.script.config.processor = processor;
+  }
+
+  // prepare files for potential consumption by request.js:
+  if (opts.script.config.files) {
+    let absoluteScriptPath = path.resolve(process.cwd(), opts.options.scriptPath);
+
+    const files = Object.keys(opts.script.config.files).reduce((fileMap, name) => {
+      let relativeFilePath = opts.script.config.files[name];
+      let filePath = path.resolve(path.dirname(absoluteScriptPath), relativeFilePath);
+      let fileStream = fs.createReadStream(filePath);
+      return Object.assign(fileMap, { [name]: fileStream });
+    }, {});
+
+    opts.script.config.files = files;
   }
 
   var runner = createRunner(opts.script, opts.payload, opts.options);
--- a/test/scripts/files.json
+++ b/test/scripts/files.json
@@ -1,0 +1,13 @@
+{
+  "config": {
+      "target": "http://127.0.0.1:3003",
+      "files": ["./pets.csv"]
+  },
+  "scenarios": [
+    {
+      "flow": [
+        {"post": {"url": "/pets", "body": "@./pets.csv"}}
+      ]
+    }
+  ]
+}
--- a/test/test.sh
+++ b/test/test.sh
@@ -73,6 +73,11 @@
   [ $? -eq 0 ]
 }
 
+@test "Run a script with file upload" {
+  ./bin/artillery run ./test/scripts/files.json | grep 'all scenarios completed'
+  [ $? -eq 0 ]
+}
+
 @test "Run a script using default options (output)" {
   ./bin/artillery run ./test/scripts/hello.json | grep "Log file: artillery_report_*"
   [ $? -eq 0 ]


hassy:
Sweet 👍

dallasvaughan:
The reference in the `post` entry doesn't seem to be using the referenced file (i.e., it uses the literal string `"@./path/to/file"` instead of resolving the file declared in the `config`).

zcei:
I think @hassy didn't release a new version, so by now `1.5.2` is not using this feature - you'll have to use `shoreditch-ops/artillery#master` or wait until a `1.6.0` is released.

dallasvaughan:
I downloaded the artillery-master and artillery-core-master source (with his changes) and attempted to run artillery locally, but it didn't find the reference in the request:

```
config:
  target: "http://localhost:8080/"
  files:
    - "./test.png"

# [SNIP]

    - post:
          url: "/myurl"
          headers: {"Content-type" : "image/png"}
          body: "@./test.png"
```

When I run with DEBUG on the result request looks like:

```
http request: {
  "url": "http://localhost:8080/",
  "method": "POST",
  "headers": {
    "user-agent": "artillery 2.4.1 (https://artillery.io)",
    "content-type": "image/png"
  },
  "body": "@./test.png"
}
```
Perhaps I have something misconfigured.

I ran the unit tests (including the new file-based ones) defined in artillery but while it succeeds it doesn't verify that it's using the content of the file rather than the literal "@./pets.csv" string.

zcei:
Sorry, I forgot that `master` is using `artillery-core@2.4.1`, which does not include my work for file upload.

So currently you'd need to manually link both dependencies together - which is not a really viable solution, but the only feasible right now. 

As @hassy is actively working on the project right now I assume he's just waiting a bit to bundle the changes into a new release 🙂

hassy:
Yeah there will be a release in the next week or so. :shipit:

git-checkout-crytek:
Hi there,

Is there a way we can pass reference to json file in POST instead of passing the complete json body in json: {{someRandomJson}} ???

zcei:
This PR is for file uploads, so you'd just pass "the reference" (aka the ReadStream of the JSON file) to the body and have it uploaded as every other binary.

hassy:
I was going to write a test for this, but the code isn't working for me, with neither single files nor multipart form data. For single files, the contents of `body` is posted (as mentioned by someone else above) and for multipart form data the request aborts early.

A server expecting multipart form uploads:

```javascript
'use strict';

var multiparty = require('multiparty');
var http = require('http');

http.createServer(function(req, res) {
  if (req.url === '/upload' && req.method === 'POST') {
    var form = new multiparty.Form();

    form.parse(req, function(err, fields, files) {
      if (err) {
        console.log(err);
      }

      res.writeHead(200, {'content-type': 'text/plain'});
      res.write('received upload:\n\n');

      console.log(fields);
      console.log(files);

      res.end('ok');
    });
  }
}).listen(8123);
```

Artillery script:

```yaml
config:
  target: "http://localhost:8123"
  files:
    - "./file.txt"
  phases:
    - duration: 1
      arrivalCount: 1
scenarios:
  - flow:
      - post:
          url: "/upload"
          #body: "@./file.txt"

          formData:
            foo: "@./file.txt"
```

hassy:
@zcei I've reverted the changes before publishing a new release of Artillery. Keen to get file upload support in though if you're still interested in working on the feature.

zcei:
Yup, actually are already looking at it - there was a small flaw in how I created the file references object (used array key instead of relative file path), but now I'm stuck with passing the streams around, they seem to never end.

My guess is that they already start emitting data, but even pausing them immediately after creation doesn't change behaviour. Is this maybe due to forking a new process? 🤔 

(If I recreate the streams right before the `request` call it works)

dstibrany:
FWIW these two changes worked for me:
```
diff --git a/lib/engine_util.js b/lib/engine_util.js
index 0e0d0e9..539dbf6 100644
--- a/lib/engine_util.js
+++ b/lib/engine_util.js
@@ -5,6 +5,7 @@
 'use strict';

 const async = require('async');
+const fs = require('fs');
 const debug = require('debug')('engine_util');
 const traverse = require('traverse');
 const esprima = require('esprima');
@@ -454,7 +455,7 @@ function fileReference(body, config) {
     return null;
   }

-  const reference = config.files[relative];
+  const reference = fs.createReadStream(config.files[relative].path);

   if (!reference) {
     return null;
```

```
diff --git a/lib/engine_http.js b/lib/engine_http.js
index a83c9fd..3960f64 100644
--- a/lib/engine_http.js
+++ b/lib/engine_http.js
@@ -182,6 +182,9 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
           if (reference) {
             acc[k] = reference;
           }
+          else {
+            acc[k] = v;
+          }

           return acc;
         },
```

And then I references files in the yaml file like:
```
files:
    "../sample_files/sample.pdf": "../sample_files/sample.pdf"
```

zcei:
While this is working, it's the same as:
> (If I recreate the streams right before the request call it works)

From a design perspective I didn't want to include `fs` module in the engine utils, as in my point of view the engine should get something usable passed instead of creating its data from just a file path (e.g. you could potentially pass in different streams to the engine and not rely on the cli implementation)

That said, I sadly hat no time further looking into how to fix this 🙁

GudarJs:
Sorry, is this implemented in 1.6.0? I'm trying but file doesn't upload :(.

hassy:
@GudarJs no, the new functionality wasn't working 100% (see the messages in this thread) and the PR hasn't been picked up since.

PR #269, opened 2017-03-08
Adds the URL to terminal output

ngbravo:
Solves https://github.com/shoreditch-ops/artillery/issues/221 by adding the URL as an option when calling printReport. 

This does **not** add the url to the report itself, just to the print.

Question: why is the URL not in the report itself?

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -272,7 +272,14 @@ For more details see: https://github.com/nodejs/node/issues/9756
         log('Report for the previous %ss @ %s',
             result.script.config.statsInterval,
             report.timestamp);
-        printReport(report, {target: result.script.config.target});
+
+        let target = result.script.config.target;
+        if (options.environment) {
+          const environments = result.script.config.environments;
+          target = environments[options.environment].target;
+        }
+
+        printReport(report, {target: target});
         spinnerOn();
       });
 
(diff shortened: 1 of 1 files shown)


hassy:
Thanks @ngbravo. I don't think this will work if we have any environments specified (via `config.environments`) though.

ngbravo:
You're right. I have a couple of questions before moving forward:: 

1. Considering there seems to be no restriction on name or amount of environments, which should be printed out?
2. Is my approach of have the URL printed out, but not in the report itself, ok for you?

Cheers!

hassy:
The user would expect the URL that the tests are running against to be printed. It's fine to only print the target in the console.

ngbravo:
@hassy Now it should work when there is an environment stated, else it defaults to the specified global target.

hassy:
Closing for now as it needs a rebase. Happy to merge if/when fixed though.

ngbravo:
I genuinely wanted to contribute and help maintain artillery, but taking 2 months to approve, and then rejecting the PR because it needs a rebase (no surprise) is quite disrespectful.

I will no longer use, recommend, or try to contribute to this project.

hassy:
I appreciate that and thanks for your time. Not that you probably care at this point, but maintaining a free open source project is a big time investment, and PRs can and do fall by the wayside unless critical and/or the contributor is active.

PR #277, opened 2017-03-13
add callback to `run`

HarryWeppner:
Option to provide callback when running `artillery` from another library.

In our case, we're generating `artillery` configuration automatically and when calling `run` require a basic `callback` mechanism to know that testing completed.

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -23,7 +23,7 @@ const createRunner = require('../runner');
 
 module.exports = run;
 
-function run(scriptPath, options) {
+function run(scriptPath, options, runCallback) {
   let logfile;
 
   // is the destination a directory that exists?
@@ -309,6 +309,10 @@ For more details see: https://github.com/nodejs/node/issues/9756
               }
             }
           });
+        }
+
+        if (runCallback) {
+          runCallback();
         }
       });
 


hassy:
Thanks for the PR @HarryWeppner! I'm in two minds about this one. Artillery should either provide a proper programmatic interface rather than the half-hearted attempt at one we currently have, or make the call to remove the ability to `require('artillery')` altogether and instead encourage users to use it as they would any other command-line tool (i.e. wrap it in a shell script for example). Thoughts?

HarryWeppner:
Hi @hassy I think there's _huge_ value of using `artillery` as a library. You can't (shouldn't) know the context in which it is used and forcing its usage through a shell is far from ideal.

In our case we're automatically generating configurations and post-processing results. For the latter, we need to know affirmatively that all scenarios ended.

I'd be very much in favor of evolving towards a programmatic interface of which its usage in a shell script is just one special case. I don't necessarily see this pull request in conflict of that direction - it simply plugs the biggest gap ;-)

HarryWeppner:
@hassy ping.

hassy:
My apologies for letting this PR linger for so long @HarryWeppner and thanks for the reminder. I don't think this is something we want to support. I understand that it may make your particular way of using Artillery more difficult, but there's a trade-off between supporting every single use-case and having a focused product which is maintainable in the long term.

Artillery is meant to be used as a CLI tool. It *happens* to be written in Node.js (because Node.js is an amazing platform), but that does not make it a library designed to be used from other Node.js modules (though there's `artillery-core`).

Out of curiosity, why not generate configurations on-the-fly, write them to a YAML/JSON file and then call the `artillery` CLI on those? (and process the JSON log) That's a very common pattern of interacting with CLI tools. You could still write everything in Node.js if you don't want to introduce shell scripts into your codebase too.

PR #302, opened 2017-05-01
Added capability to accept js files as input.

### LINKED ISSUE #301: Added capability to accept .js files in addition to .json files.
Using nodejs module syntax allows me to do scenario sharing, config sharing etc. that can't be done via json. I plan on supplying a pull request with this feature enabled.

schwer:
closes #301


ericmacfarland:
@hassy Would you mind taking a look at this pull req? The ability to write script files in js adds considerable flexibility and power for the user :)

hassy:
I'd really like to see a use case for it. I'm not really seeing the benefit of being able to write some JSON and export it from a JS module, over just sticking it in a .json (or better yet a YAML) file. I can see quite a bit of potential for confusion though (e.g. trying to define a processor function inline in the js file - that's not going to work).

ericmacfarland:
If the user is simply writing JSON, then you are correct that storing it in a `.json` file would be the preferred option.

However, if the user wants to perform dynamic operations at run-time, they currently have very limited options.

A script written in JS can be `require`ed the same way a script written in JSON can, so the JS script would just have to export a config with the same schema as the JSON script. The user could still easily define a processor function with the built in functionality.

I'll share a use case once I polish it a bit.

ericmacfarland:
## Example use case

### Sharing environment configs between scripts
#### Background:
1. Your application has three environments (`dev`, `stage`, `prod`), each with their own address
2. The services inside each env have the same base address `/api`
3. You're a diligent performance tester that has created 50 scripts to test your app

#### The issue
Now suppose that a new feature necessitates that the base address change from `/api` to `/api/v1`

#### In the current state
You have to change the base address in 3 places for each of the 50 files (150 changes)

Your best option in this case is probably to do a search & replace for the address, taking care to not replace any unrelated occurrences of `/api`

#### If script files can be written in JS
You previously realized that the same config is used in many places, and simply stored the addresses in a single file that each script would `require`.

You easily make the 3 address changes in the config file, and all scripts utilize the new config.

@hassy

ericmacfarland:
### Handling authentication with tokens (e.x. OAuth)
#### Background
1. Your application is behind authentication that requires tokens to be inluded in the request header
2. The tokens expire after a period of time, say an hour
3. You already have a script that can fetch new tokens and store them in JSON files

#### The issue
All of your scripts have to get the tokens somehow

#### In the current state
There are two options:
1. Have the token inside a `.csv` payload file
    - Requires the extra step of writing the token from the `.json` file to a new `.csv` file
    - If you have multple payload files, each one must have the token in it, and you must have a way to update the tokens in that file
    - Note: payload files seem more geared for static data, not dynamic data, so this feels like hammering a nail with a wrench (in my opinion)
2. Write a `beforeRequest` processor function that can `require` the token file and insert the tokens into the request
    - Every request must now call this processor function (an extra line for each request, in each script)
    - It's less apparent in the script that you're handling the authentication, it would be more succinct to just specify a header

If you have multiple environments that require different tokens:
- The payload method becomes considerably more complex and gets messy.
- The processor function can handle this with some extra code, it just needs to check which environment is being used and know where to look for the tokens

#### If script files can also be written in JS
The script file could dynamically add in-line variables to each environment's config, and include the header in each request by simply specifying a default header (existing functionality)

hassy:
Thanks for taking the time to write these up @ericmacfarland. Use case no. 1 I think is solved with the `--config` option in `run`. Use case no. 2 can be handled in a couple of ways:

1. As an explicit request at the beginning of the scenario which captures the auth token. Subsequent requests can reuse that token then.
2. If each request needs to acquire a token, `beforeRequest` hooks can be set at scenario level (and will run for each request).
3. Using a `function` action at the beginning of the scenario to acquire the token and insert it into `userContext.vars` so that subsequent requests can reuse it.

ericmacfa:
I can't seem to find the `--config` option in the [cli reference](https://artillery.io/docs/cli_reference.html), could you please share some information on that option?

I also wasn't able to find any docs around adding a `function` action to the scenario, do you have any examples of this?

hassy:
Ah yes sorry, the docs are missing some recent stuff. I've updated them now:

1. `--config`: https://artillery.io/docs/cli_reference.html#run
2. `function` action: https://artillery.io/docs/testing_http.html#the-function-action

ericmacfa:
@hassy Thanks for the docs! I have a few questions, but I'll create a separate issue for those instead of cluttering the pull req thread

schwer:
I opened the issue because it greatly simplified the what was doing with artillery.

I was using artillery to load test a web service with a complex input structure. The config option was insufficient to carry the input structure. The input structure needed to be shared amongst a variety of tools I use for testing. A simple require statement was the most obvious solution for the predicament.

schwer:
As time passes, I'm finding it harder and harder to remember the nuances of why I created Issue #301. I remember constructing the artillery load test and realizing that I would either need to add a pre-processing step to serialize my list of complex web-service payloads into JSON to supply them to the config file; or I would need to add support for javascript config files. Having worked with webpack in the past, I saw the advantage and relative simplicity of allowing configuration to be supplied via as javascript.

Is there a timeframe after which I can expect this PR to be approved or closed?

hassy:
@schwer Would custom JS functions have solved your problem?

https://artillery.io/docs/http-reference/#advanced-writing-custom-logic-in-javascript

As it stands, I'm not seeing how the benefits of being able to write an Artillery test as a JS object inside a '.js' file outweigh the added complexity (new docs, new tests, potential for confusion, inability to just `JSON.parse` or `YAML.load` a script). Happy to revisit the decision though, it's not final.

schwer:
Now that I've found the time, I've added some comments to #301 that I hope might add some more color to this PR.

PR #338, opened 2017-08-04
Cannot set v8debug in modern node

### LINKED ISSUE #337: v8debug never true in node 8, so child debug port never set.
In nodejs 8 we now have the --inspect flag that hooks up with Chrome devtools. The option to expose debug as v8debug is gone.

`node: bad option: --expose_debug_as=v8debug`

While the docs are OK, there is still a need to read through the source and to be able to step through the code.

The debug-mode module on npm seems to use a 'dirty check' by trying to use the debug port. If it fails, then it assumes it is debugging. It also has the `v8debug` check.

plumpNation:
When trying to debug with the settings expected by the code one receives
`node: bad option: --expose_debug_as=v8debug`.

Without that option, v8debug will never be true, and the debugger will
never be attached to the child process.

I don't think this solution will be 'enough', as it uses the `--inspect` flag to start up the child process and that may not be backward compatible enough.

It also introduces another dependency, [debug-mode](https://www.npmjs.com/package/debug-mode) by IonicaBizau.

Fixes #337

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -13,6 +13,7 @@ const os = require('os');
 const path = require('path');
 const EventEmitter = require('events');
 const fork = require('child_process').fork;
+const debugMode = require('debug-mode');
 
 const debug = require('debug')('artillery:runner');
 const L = require('lodash');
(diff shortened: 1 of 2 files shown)


hassy:
Thanks for the patch @plumpNation! I'm thinking we might just look for an `ARTILLERY_DEBUG` environment variable (or some such) and enable `--inspect` if that's set? The benefits I see are:

- We'd avoid an extra dependency (which based on my cursory research seems to use undocumented APIs and isn't guaranteed to be future-proof)
- Debugging would still work under older versions of node

As-is, we're already requiring for an extra option to be set (`--expose_debug_as=v8debug`), so we might as well replace that with an environment variable.

Thoughts?

plumpNation:
Agreed, shouldn't really add the extra dependency but it was so small.

I'm sure that your fix will certainly please more users than mine due to the backward compat issue.

Thanks for the awesome tool, really looking forward to it's growth.

PR #377, opened 2017-09-26
Allows for csv-parse options to be passed via options property on payload.

gwsii:
@erikerikson can you have a quick look?

As some of our payload files contain data with commas, which are passed to the endpoint on the URL, this is an expedient way to make the payload processing in Artillery more flexible.

Verified this will work for us.

Thanks,

~Greg

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -332,7 +332,8 @@ function readPayload(script, callback) {
     script.config.payload,
     function readPayloadFile(payloadSpec, next) {
       let data = fs.readFileSync(payloadSpec.path, 'utf-8');
-      csv(data, function(err, parsedData) {
+      const csvOpts = payloadSpec.options || {};
+      csv(data, csvOpts, function(err, parsedData) {
         payloadSpec.data = parsedData;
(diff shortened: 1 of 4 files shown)


lib/commands/run.js
```diff
@@ -332,7 +332,8 @@ function readPayload(script, callback) {
     script.config.payload,
     function readPayloadFile(payloadSpec, next) {
       let data = fs.readFileSync(payloadSpec.path, 'utf-8');
-      csv(data, function(err, parsedData) {
+      const csvOpts = Object.assign({}, payloadSpec.options);
```
erikerikson: it might be more idiomatic to write this as:
`const csvOpts = payloadSpec.options || {};`

hassy:
Thanks @gwsii! LGTM, but may I request a commit message change: a prefix + description in the imperative < 72 characters. Something like `feat: Allow for options to be passed through via config.payload.options`.

([commitlint](https://github.com/marionebl/commitlint) really needs to be part of the `precommit` hooks, just haven't got round to it yet)

PR #381, opened 2017-09-27
feat: Allow for options to be passed through via config.payload.options

gwsii:
NOTE: Re-opening after rebasing. Addresses feedback from other PR.

As some of our payload files contain data with commas, which are passed to the endpoint on the URL, this is an expedient way to make the payload processing in Artillery more flexible.

Verified this will work for us.

Thanks,

~Greg


hassy:
Nice one, thanks @gwsii!

PR #446, opened 2018-03-25
feat: Allow for scenario variables to be passed in with "-v" option

hassy:
Add "-v" option to "run" command to allow for scenario variables
to be set at run time, for example:

```
artillery run -v color=red,height=120 test.yaml
```

Makes variables "color" and "height" available in the script (with
the respective values of "red" and "120").


alexgocariq:
I am looking for the similar need where I need to provide 
        - duration: {{duration}}
          arrivalRate: 30
          rampTo: 30
          maxVusers: 30
I was using -v duration=10 but it did not work what is the correct usage?

PR #464, opened 2018-05-08
feat: Allow for scenario variables to be passed in with "-v" option

hassy:
Add "-v" option to "run" command to allow for scenario variables
to be set at run time using JSON, for example:

artillery run -v '{"color":"red", "height":120}' test.yaml

Makes variables "color" and "height" available in the script (with
the respective values of "red" and "120").


hassy:
Fixes https://github.com/shoreditch-ops/artillery/issues/454

PR #482, opened 2018-05-31
Memory leakage + upgrade packages with vulnerabilities

DY6229:
We had a memory leakage because of data that is being saved that seemed to be unnecessary. 
(The _entries that is being saved in the aggregate object)
We deleted the data and also upgraded the 'lodash' and 'debug' packages because of known vulnerabilities issues:
https://nodesecurity.io/advisories/534
https://nodesecurity.io/advisories/577


hassy:
Thanks for the PR @DY6229! I'll try to review & respond asap.

enudler:
Very useful fix

enudler:
Added another commit.
This commit has the option to make requests before test start to run. the artillery json now can have another section called 'before', which has a  single flow (like in a scenario) that will run once before the test. 
also now this fork is rebased to this project with the last few merges.

DY6229:
Added another commit.
This commit has the option to make requests before test start to run. the artillery json now can have another section called ‘before’, which has a single flow (like in a scenario) that will run once before the test.
also now this fork is rebased to this project with the last few merges.

Kosta-Github:
@DY6229 is the `before` section only run `once` globally or run `once` for each individual `virtual user`? I would be in need for the latter, e.g., for a user login flow to generate an `auth` token used for that user for all subsequent requests...

@hassy any chance to get that in?

hassy:
@Kosta-Github you can use a `function` step in your scenario definition to set up some state for each new virtual user: https://artillery.io/docs/http-reference/#function-steps-in-a-flow

@DY6229 There are 3 separate changes in this PR - is there any chance you could split those out (+ investigate why tests are failing)? The new `before` hook especially would be extremely useful. :)

NivLipetz:
@hassy, @Kosta-Github the before section is in PR here: https://github.com/artilleryio/artillery/pull/612

PR #490, opened 2018-06-18
fixes #487

### LINKED ISSUE #487: package "open" has a critical security vulnerability
package is deprecated anyway, so maybe best course of action would be to swap it for https://github.com/sindresorhus/opn

capaj:
use opn instead of open package, fixes #487


hassy:
Thanks @capaj! Would you mind removing `package-lock.json` from the PR? That file will be generated by the CI server before publishing the package.

PR #513, opened 2018-08-09
Better support for in-process use of Artillery run command

gwsii:
TL;DR: Provide script and return results in-memory to run command. Improve logging and plug memory leaks. 

Enables invoking Artillery run command directly in-process via passing the script as a parameter (rather than a file path) into the Run command.

Further supports in-process use by allowing results to be returned via a callback, made available as `options.output`.

These changes also address memory leaks caused by uncancelled event handlers, including the SIGINT, SIGTERM handlers and the CPU measurement feature interval handler. Also makes an improvement to debug logging.

Added appropriate tests and verified eslint is error free.

Thanks for your consideration!


hassy:
Holy moly, time flies! I'm sorry I haven' got round to reviewing this PR properly yet @gwsii! I'll try to get it done in the next couple of days.

To quickly jot down some general thoughts  - there have been PRs with changes along similar lines previously (trying to make Artillery easier to use as a library), but I'm not sure if that's a use-case we really want to support long term. I personally lean towards making the CLI easier to use from any other language (whether it's bash, Python, or indeed Javascript) by following *nix best-practices and providing integration points via e.g. JSON output on `stdout`. This is no way a final position of course, and is definitely something up for discussion.

PR #514, opened 2018-08-10
dep: Switch from rolex to driftless

### LINKED ISSUE #500: Dependency rolex was deprecated and renamed to driftless
When installing artillery, npm shows a deprecation notice for the dependnecy [rolex](https://www.npmjs.com/package/rolex). It was renamed to [`driftless`](https://github.com/dbkaplun/driftless) and is also available in a newer version 2.0.3.

I think this should be updated and changed to avoid cluttering users with deprecation notes that they cannot influence directly.

lzhang:
Fixes #500.


hassy:
Thanks @lzhang!

PR #517, opened 2018-08-13
Add whileTrue condition in WebSocket and Socket.io engines.

### LINKED ISSUE #468: Add WhileTrue to other engines
It seems like the `whileTrue` loop conditional is only supported by http. Can we add it to the other engines? Happy to do a PR, just not sure if there is reasonings behind not including it, other than it's experimental?

shulard:
Since the logic in actually handled by the `utils` engine, I've just added the `whileTrue` property in step definition.

Must fix #468.


hassy:
Thanks @shulard!

Issue #510, opened 2018-08-02
Support for processor in WS engine


patilarpith:
Posted this on Gitter to no avail. 

I have a slightly complex message structure which I'm passing via a processor function but I see a crash at https://github.com/shoreditch-ops/artillery/blob/master/core/lib/engine_ws.js#L70
with message:
TypeError: Cannot read property 'toString' of undefined

So I was wondering if my configuration file was incorrect or support for processor in WS engine is not available yet

```yaml
config:
  target: "ws://localhost:8000/"
  phases:
    - duration: 5
      arrivalRate: 5
  ws:
    rejectUnauthorized: false
  processor: "./generator.js"
scenarios:
  - engine: "ws"
    flow:
      - function: "generateData"
      - send: "{{ message1 }}"
      - think: 2
      - send: "{{ message2 }}"
      - send: "{{ message3 }}"
      - think: 5
```

patilarpith:
@hassy Thoughts?

hassy:
Custom functions are supported in the WS engine. I would guess (without seeing what `generateData` does) that `message1` is not being set in the custom function. Can you also confirm that you're running the latest version of Artillery? (`1.6.0-22`)

patilarpith:
@hassy Yes, I can confirm that I'm using `1.6.0-22`

Here is my test generator.js file
```javascript
module.exports = {
  generateData
};

function generateData(context, events, done) {
  context.vars['message1'] = 'hello';
  context.vars['message2'] = 'welcome';
  context.vars['message3'] = 'greetings';
  console.log(context.vars);
  return done();
}
```
Commenting out  `- engine: "ws"` line seem to run the scenarios loop but the moment I set engine to "ws" it crashes with

```
TypeError: Cannot read property 'toString' of undefined
    at f (<Absolute path>/node_modules/artillery/core/lib/engine_ws.js:70:25)
    at fn (<Absolute path>/node_modules/artillery/node_modules/async/lib/async.js:746:34)
```

patilarpith:
@hassy Did you get a chance to to look into this?

tamanugi:
I got a same error. 

I send a PR for fix this issue.
https://github.com/shoreditch-ops/artillery/pull/527

I'm glad if you can check it

PR #527, opened 2018-08-27
fix: #510 TypeError occurs with custom function in WS engine

tamanugi:
this request is to fix issue #510


hassy:
Thanks @tamanugi!

PR #562, opened 2018-11-09
fixes #561 - enable zero value assertions

### LINKED ISSUE #561: Zero values not being matched in expect equal function
👋 Hey!

When using `artillery-plugin-expect`, comparing against zero values causes an error,

#### Object to evaluate:
```js
{
  valueThatEqualsZero: 0
}
```

#### Artillery assertion:
```yml
- equals:
  - "{{ valueThatEqualsZero }}"
  - "0"
```

#### Outcome
```
not ok equals , 0
```

### Cause
It looks like it's being caused by [this line](https://github.com/shoreditch-ops/artillery/blob/master/core/lib/engine_util.js#L257). As if the value is falsey, it will return an empty string.

JoeScho:
👋Hey stranger! This a fix for #561 - in which checking a value is equal to 0 isn't working as expected. 

Any problems, just give me a shout!

### DIFF
--- a/core/lib/engine_util.js
+++ b/core/lib/engine_util.js
@@ -254,7 +254,7 @@ function renderVariables (str, vars) {
     if (matches[0] === str) {
       // there's nothing else in the template but the variable
       const varName = str.replace(/{/g, '').replace(/}/g, '').trim();
-      return L.get(vars, varName) || '';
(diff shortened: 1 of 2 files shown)


hassy:
Thanks @JoeScho!

Issue #564, opened 2018-11-12
[Feature Request] ifTrue: Allow Processor Functions


kush-josh:
Hi,

I looked at artillery codebase, and it seems (according to my understanding) the best way to skip a function in the flow is to use undocumented `ifTrue`. 

However, I am wondering whether it is possible to allow ifTrue to use Processor functions? 
If that is not possible, is it possible to use `beforeRequest` to signal Artillery to skip this request? (This would be preferred way, if possible)

hassy:
That sounds like a good feature to have @kush-josh! There's no way to skip a request with a `beforeRequest` function at the moment, one can only abort the entire virtual user scenario by calling the callback with an `Error`.

kush-josh:
I am going with enhancing of `beforeRequest` to allow values to be skipped since `ifTrue` changes are not that easy, and would break backward compatibility.

So, to use `beforeRequest` to skip the scenario, you need to throw errors which has `ignore =  true` set.

Can do this by:

```javascript
function sample1(requestParams, context, ee, next) {
     return next({ignore: true})
}
```

```javascript
class SkipError extends Error {
    constructor(message) {
        super(message);
        this.ignore = true;
    }
}

function sample2(requestParams, context, ee, next) {
     return next(throw SkipError("Sample Error"))
}
```

I am not sure whether you want me to update the docs since `ifTrue` was an undocumented feature. I have erred on side of caution and left it for now.

PR #566, opened 2018-11-12
Allows skipping of request in beforeRequest - #564

kush-josh:
Added a check in Artillery HTTP Engine to check for ignore value
If the check is true, then Artillery will proceed with next request
This allows invalid requests to be skipped rather than exit scenario

### DIFF
--- a/core/lib/engine_http.js
+++ b/core/lib/engine_http.js
@@ -189,12 +189,12 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
       let cond;
       let result;
       try {
-        cond = filtrex(params.ifTrue);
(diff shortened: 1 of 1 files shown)


hassy:
Thanks for the PR @kush-josh! It doesn't feel right to me to overload the meaning of returning an error in that way (and having to extend one's error object as well). I think the right way to go is to either find of a way to make it work with `ifTrue`, or perhaps to allow for options to be passed to the `beforeRequest` callback.

kush-josh:
Usage:

processors.js
```javascript
function sampleCondition(contextVars) {
    return false;
}
```

artillery.yaml
```yaml
- get:
    ifTrue: sampleCondition       # Will execute from processor
    url: foo/bar
- get:
   ifTrue: 1 <  5
   url: foo/bar/baz                     # Will execute statement directly
```

hassy:
Nice, thanks @kush-josh!

PR #577, opened 2018-11-25
Fix/sanitise false value

aruiz-caritsqa:
- false value is now sanitised the same way as "0"


hassy:
Nice, thanks @aruiz-caritsqa!

PR #580, opened 2018-11-27
feat: Add extra options for config.payload

hassy:
Add `skipEmptyLines`, `cast`, `skipHeaders` and `delimiter` options.

- `skipEmptyLines` (default `true`) -- skip any empty lines in the CSV file
- `cast` (default `true`) -- convert values to native types (e.g. numbers)
- `skipHeader` (default `false`) -- skip the header row when populating variables
- `delimiter` (default `,`) -- set the delimiter


rimantoro:
Hi,

Do you guys already release this commit ?

I'm trying to use ```cast``` option to ```true``` but the expected parameter still send as string instead of number.

Issue #205, opened 2016-11-11
count down  (rampTo from big to small) don't works 


servusoft:
counting up works fine:
```
    - duration: 60
      arrivalRate: 100
      rampTo: 200
``` 
but count down break the phase and exit the test early
```
    - duration: 60
      arrivalRate: 200
      rampTo: 100
```

hassy:
`rampTo` was not intended to be used to ramp down. I'll take a PR that fixes that. :)

servusoft:
would be better split `rampTo` to `rampUp` and `rampDown` or should rampTo 'understand' both cases ?

PR #606, opened 2018-12-16
feat: Support ramp-down as well as ramp-up

joeattardi:
The rampTo option can now also be set to a value lower than arrivalRate, and it will ramp from the starting arrivalRate down to the rampTo value.

Implements #205


hassy:
Nice one @joeattardi :)

PR #614, opened 2018-12-25
feat: Add request path to response entries

NivLipetz:
Adds the request path to each response entry. This feature is useful when looking at the test as a whole, or at periods of time, to narrow down which requests returned 2xx/4xx/5xx as well as their latencies. This enhancement of metrics allows for better understanding of each endpoint the test loads.

### DIFF
--- a/core/lib/engine_http.js
+++ b/core/lib/engine_http.js
@@ -256,6 +256,7 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
           requestParams.uri = template(requestParams.uri, context);
         }
         if (requestParams.url) {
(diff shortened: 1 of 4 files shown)


hassy:
This could be done with a custom `afterResponse` JS function rather than by extending the internal stats object. Any reason that wouldn't work for your use-case?

NivLipetz:
For my use-case it doesn't seem that I can recreate my needed behaviour by only using the `afterResponse` JS function.

The motivation of this feature is to be able to audit the number of requests sent and response codes received from each distinct endpoint. This can only be done by adding the path to _each_ entry as done in the file `/core/lib/runner.js` where the event `response` is listened to.

PR #615, opened 2018-12-25
feat: Add the ability to limit the number of concurrent requests

NivLipetz:
Load tests that are too intensive for the service they are testing will overwork the CPU of the machine running the test because of a build up of requests that should have been sent but have not. This will cause the test to have inaccurate results in the report and to be irrelevant.

This feature allows the user to configure a limit of requests that can be pending by "throwing away" requests when this limit has been reached. Every time a pending request is completed, another spot is freed up in the pool for another request to be sent. In order to keep track of the requests that were thrown away, the report returns a scenariosAvoided counter.


hassy:
Thanks @NivLipetz! I think it's worth clarifying that this change allows for a limit to be set for max virtual users rather than requests. Scenarios which send a lot of requests (e.g. with a `loop`) could still overwhelm the target service.

I'm thinking the way this should work is as an option on phase definitions, e.g.:

```yaml
config:
  target: ws://my-service.dev
  phases:
    duration: 600
    arrivalRate: 10
    maxVusers: 200
```

Thoughts?

.travis.yml
```diff
@@ -1,22 +0,0 @@
-language: node_js
```
hassy: We don't want to remove this file. Were the tests failing on Travis?

NivLipetz: No problem, returned it

NivLipetz:
Implemented as you suggested with the `maxVusers` in the phase configuration. This will limit the number of virtual users running concurrently if configured.

PR #619, opened 2018-12-27
feat: Add custom functions that can be called inside test script

NivLipetz:
This feature enables dynamically adding functions to the contextFuncs object in core/lib/runner, which are functions that the Runner knows how to execute if they are called in the Artillery script. Currently, the only functions there are $randomString and $randomNumber, which might not be enough for some tests. We, for example, needed to randomly generate uuid and dates, and therefore needed new functions that support this. By having this feature, we could easily add these functions (and more if needed) to our test.

An example of usage can be seen in the test provided.

### DIFF
--- a/core/lib/runner.js
+++ b/core/lib/runner.js
@@ -18,6 +18,7 @@ const createPhaser = require('./phases');
 const createReader = require('./readers');
 const engineUtil = require('./engine_util');
 const wl = require('./weighted-pick');
+const {$randomNumber, $randomString, $uuid, $dateNow} = require('./template-strings');
 
 const Engines = {
   http: {},
@@ -35,7 +36,9 @@ module.exports = {
   stats: Stats,
   contextFuncs: {
     $randomString,
-    $randomNumber
+    $randomNumber,
+    $uuid,
+    $dateNow
   }
 };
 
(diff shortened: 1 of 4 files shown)


hassy:
I'm hesitant to introduce another way to load custom JS code; we should be able to reuse the existing `config.processor` option to provide additional functions for template strings.

Another thought is that perhaps the functions you need should just be added alongside the default ones, as they're likely to be something that others may want to use.

NivLipetz:
I agree with you regarding adding another way to load custom JS code. I hard-coded the functions we needed and refactored out the templates to an external file, that way more templates can be added easily.

SandroGrzicic:
I looked at this PR which seems to add custom functions support, but presumably the documentation for it hasn't been written yet, so can you please provide a short example of how to use custom functions, for example within the "config.target" property (to generate a dynamic query string parameter value)? Thanks.

NivLipetz:
Hi @SandroGrzicic sorry for the late reply.. you simply add `()` to the function call like this: `$myCustomRandomIdGeneratingFunction()` 
```
config:
  target: "wss://my.websocket.host/?id={{ $myCustomRandomIdGeneratingFunction() }} "
```

NivLipetz:
@hassy any update about this PR?

PR #623, opened 2019-01-02
Fix to start csv-parse from_line parameter from 1.

### LINKED ISSUE #622: Regression in CSV parsing because of csv-parse upgrade.
A new version of `csv-parse`, i.e. `4.3.0` was released on Dec 31, 2018. This version has some validation refinements including one for the `from_line` option. artillery passes the `from_line` option as either 1 or 0 depending on whether headers should be skipped or not. This should be changed to use 2 or 1 instead.

Affected line in `artillery/lib/commands/run.js`:

    from_line: typeof payloadSpec.skipHeader === true ? 1 : 0,

This should be changed to:

    from_line: typeof payloadSpec.skipHeader === true ? 2 : 1,

Alternatively, package.json should be changed to use 

    "csv-parse": "4.2.0",

rajeshja:
CSV parsing was breaking with version 4.3.0 of csv-parse, which introduced some validation refinements. Changing the from_line parameter to 1 when not skipping headers, and 2 when skipping headers fixes the problem. Fixes #622

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -249,7 +249,7 @@ function readPayload(script, callback) {
       let csvOpts = {
         skip_empty_lines: typeof payloadSpec.skipEmptyLines === 'undefined' ? true : payloadSpec.skipEmptyLines,
         cast: typeof payloadSpec.cast === 'undefined' ? true : payloadSpec.cast,
-        from_line: typeof payloadSpec.skipHeader === true ? 1 : 0,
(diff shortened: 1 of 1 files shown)


rajeshja:
I can't tell how to fix the circleci issue. If someone can point that out, I can make the appropriate changes.

NivLipetz:
in order to fix the circleci issue you need to change your commit message since it is failing on commitlint.

You can read more about it here: https://github.com/marionebl/commitlint

hassy:
Nice, thanks @rajeshja

PR #624, opened 2019-01-09
fix: Memory leak

NivLipetz:
This merge request solves a memory leak by removing the _entries object saved on each intermediate object after every 'stats' emitted event.

The entry data object, which was created for each response as an array with 4 data fields, was congesting the heap. Tests running for 30 min+ with heavy load were crashing because the memory leak was stressing the CPU.

This fix enables tests under heavy load run for a long time without crashing, while still keeping all of the test statistics and measurements.


hassy:
Nice, thanks @NivLipetz!

PR #625, opened 2019-01-09
feat: Run a before test flow

NivLipetz:
Adds a "Before" section to the artillery test schema. This feature enables configuring a flow which will run once before the test and capture variables into the context that can be used in later scenarios in the test.

This is a very useful feature as it allows the user to easily add a before requests flow using the syntax of artillery, without having to load external JS files.

A relevant use-case that this feature can solve: https://github.com/artilleryio/artillery/issues/629

### DIFF
--- a/core/lib/runner.js
+++ b/core/lib/runner.js
@@ -44,7 +44,7 @@ function validate(script) {
   return validation;
 }
 
-function runner(script, payload, options, callback) {
+async function runner(script, payload, options, callback) {
   let opts = _.assign({
     periodicStats: script.config.statsInterval || 10,
     mode: script.config.mode || 'uniform'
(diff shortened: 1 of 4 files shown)


Ellesedil:
Is there any chance of this feature being merged into master? This would be extremely useful for things like generating a JWT or other one-time setup stuff (I'm using a function) that I'm not actually interested in testing because something else is responsible for it, but is needed in order to successfully test my service's performance.

I ask because it looks like this PR has been open for a while, long enough to invite some conflicts, so it's unclear if there's any internal desire to add a `before` feature like this.

guzmo:
Any news? Would be great to have this merged :)

mpatino117:
updates on this feature?

hassy:
Rebased and merged via https://github.com/artilleryio/artillery/pull/715

Thank you for a great contribution @NivLipetz, and apologies for taking so long. Time flies!

Amirsharifico:
Hi,
How can I use a variable in different *.yml file? the `before`  solution works only within one file but I need to get hold of variables in different files

PR #627, opened 2019-01-11
Allows captured values to be saved in Object attributes

aruiz-caritsqa:
- Allows you to have a `pet` object in `ctx.vars`, and set attributes of that `pet` with `capture`


aruiz-caritsqa:
commitlint fail

PR #628, opened 2019-01-12
fix: Allow capture to save to Object attributes

aruiz-caritsqa:


hassy:
Nice one, cheers @aruiz-caritsqa!

PR #635, opened 2019-01-27
 fix: Remove faulty "typeof"

RobinKnipe:
```js
from_line: typeof payloadSpec.skipHeader === true ? 2 : 1
```
The above always evaluates to `1`, as `typeof payloadSpec.skipHeader` is `boolean`, which doesn't `=== true`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,19 +19,19 @@
 
 ---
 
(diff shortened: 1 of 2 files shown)


RobinKnipe:
:+1:

PR #648, opened 2019-02-08
feat: Add headers to quick

JonForest:
I've added a `--headers` option to the `artillery quick` command.

Sorry I've not added tests; that batch file stuff is a little beyond me.  I have manually tested this with an authorisation token, and it seems to work fine.


JonForest:
Sorry about the delay in fixing up the conflict.

JonForest:
As far as I can tell, this job didn't actually fail tests?
image

Issue #537, opened 2018-09-05
I couldn't set the scenario.weight with variables


jrongl:
I just want to set the 'weight' with some variables, for example config.variables, but artillery did not work.
{
  "config": {
      "target": "https://localhost:3000",
      "phases": [
        { "duration": 10, "arrivalRate": 5 }
      ],
      "variables":[
        {
          "weight1": 1,
          "weight2": 2
        }
      ]
  },
  "scenarios": [
    {
      "weight": "{{{ weight2 }}}",
      "flow": [
        {"get": {"url": "/"}}
      ]
    },
    {
      "weight": "{{{ weight1 }}}",
      "name": "weighted scenario",
      "flow": [
        {"get": {"url": "/"}}
      ]
    }
  ]
} 
Then run this json, I got the following error:
/usr/local/lib/node_modules/artillery/core/lib/runner.js:363
        script.scenarios[i].name,
                            ^
TypeError: Cannot read property 'name' of undefined
    at runScenario (/usr/local/lib/node_modules/artillery/core/lib/runner.js:363:29)
    at EventEmitter.<anonymous> (/usr/local/lib/node_modules/artillery/core/lib/runner.js:252:5)
    at emitNone (events.js:106:13)
    at EventEmitter.emit (events.js:208:7)
    at UniformProcess.<anonymous> (/usr/local/lib/node_modules/artillery/core/lib/phases.js:181:10)
    at emitNone (events.js:106:13)
    at UniformProcess.emit (events.js:208:7)
    at /usr/local/lib/node_modules/artillery/node_modules/arrivals/lib/index.js:47:10
    at NanoTimer.setInterval (/usr/local/lib/node_modules/artillery/node_modules/arrivals/node_modules/nanotimer/lib/nanotimer.js:220:5)
    at Immediate._onImmediate (/usr/local/lib/node_modules/artillery/node_modules/arrivals/node_modules/nanotimer/lib/nanotimer.js:203:68)

PR #694, opened 2019-04-27
feat(runner): Allow scenario weights to be set with templates

hassy:
This PR adds support for setting scenario weights with variables.

Template strings that set weights may use inline variables from `config.variables` or values from the environment through `$processEnvironment`.

If a weight cannot be parsed as a number it will be set to 0.


hassy:
Fixes #537

PR #715, opened 2019-07-11
Rebase & merge #625

hassy:
Ref: #625

### DIFF
--- a/core/lib/runner.js
+++ b/core/lib/runner.js
@@ -44,7 +44,7 @@ function validate(script) {
   return validation;
 }
 
-function runner(script, payload, options, callback) {
+async function runner(script, payload, options, callback) {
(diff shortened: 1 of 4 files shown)


lexicalninja:
Is there an example of this and where it goes in the config? Thanks!

tbradley-sans:
> Is there an example of this and where it goes in the config? Thanks!

@lexicalninja : There is an example in the form of a test script, here: https://github.com/artilleryio/artillery/pull/625/commits/37bbe3b9319bf60063db32dbed369f57dd8d102f#diff-8a912a2bd0b4e992c1b821b24fa97171

It's in JSON, not YAML (ironic, when YAML is the preferred language for Artillery scripts) but if you can mentally translate from one to another, you can probably figure it out.

PR #743, opened 2019-09-12
Response codes and errors graph correction in report

onionka:
Response Codes and Errors graphs are being interpolated wrongly because there were missing 0 values. This PR is adding zeros to the places with missing values in these graphs. I had to fix it in the report manually, so I thought it would be nice to contribute to here.

This is what it looked like before fix:
Screenshot from 2019-09-12 08-38-28

### DIFF
--- a/lib/report/index.html.ejs
+++ b/lib/report/index.html.ejs
@@ -348,21 +348,27 @@ var chartData = l.map(
 //
 
 if (l.size(Report.aggregate.codes) > 0) {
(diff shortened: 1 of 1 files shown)


hassy:
Thanks @Cmiroslaf! It's very difficult to see where the change is exactly though because the entire file seems to have been reformatted. Would you mind updating the PR keeping the original whitespace so that the change can be seen?

Cmiroslaf:
Sorry for the late response. Ok, I removed it.

PR #811, opened 2020-03-05
stats: add support for artillery-engine-grpc

kenju:
Now stats2.js know gRPC engine's metrics as follows:

- engine.grpc.responses
- engine.grpc.request_rate
- engine.grpc.response_time
- engine.grpc.codes.[\d]

ref. https://github.com/kenju/artillery-engine-grpc/issues/12

---

Would you review this PR? 

I'm mentioning @hassy by having looked at https://github.com/artilleryio/artillery/blob/master/CONTRIBUTING.md#need-to-get-in-touch 📝

### DIFF
--- a/core/lib/stats2.js
+++ b/core/lib/stats2.js
@@ -148,14 +148,14 @@ Stats.prototype.report = function() {
   });
   result.scenariosCompleted = this.getCounter('core.scenarios.completed');
   result.scenariosAvoided = this.getCounter('core.scenarios.skipped');
-  result.requestsCompleted = this.getCounter('engine.http.responses')|| this.getCounter('engine.socketio.emit') || this.getCounter('engine.websocket.messages_sent');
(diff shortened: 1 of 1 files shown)


kenju:
@hassy I suppose you're too busy, but would you let me know when you are available for reviewing? Thank you in advance.

kenju:
@hassy If you don't like other engine's specification implemented here (and I agree with this feeling), we have few options...

- add option to artillery not to output metrics
  - actually, [artillery-engine-grpc output its own metrics](https://github.com/kenju/artillery-engine-grpc/pull/13) so it should be fine
- we should pass more general name, instead of passing per-engine specific names here
  - we also should fix builtin's http/socketio/websocket engines as well
- migrate grpc into artillery itself (if you have gRPC applications' usecase, I think it's nice 😉 )

kenju:
@hassy 

Btw, I'm passionate about working on any of those options 👍

kenju:
I'm closing this issue because now it is staled... 😢 

@hassy I'm happy to reopen when you would review this PR

PR #820, opened 2020-04-24
Correct documentation

vegerot:
]→)


hassy:
Thanks @vegerot! better late than never eh 😧

Issue #828, opened 2020-05-04
Header capture doesn't handle upper case characters


dandenton:
When capturing header information it seems like the header's name must be all lower case.  The following script will run the first request but it stops before running the second.

```
config:
  target: https://artillery.io
  phases:
    - duration: 1
      arrivalRate: 1
scenarios:
  - flow:
    - get:
        url: /
        capture:
          - header: ETag
            as: pageETag
        expect:
          - statusCode: 200
    - get:
        url: /
        headers:
          If-None-Match: "{{ pageETag }}"
        expect:
          - statusCode: 304
```

```
All virtual users finished
Summary report @ 12:55:37(-0600) 2020-05-04
  Scenarios launched:  1
  Scenarios completed: 0
  Requests completed:  1
  Mean response/sec: 2.17
  Response time (msec):
    min: 124.7
    max: 124.7
    median: 124.7
    p95: 124.7
    p99: 124.7
  Scenario counts:
    0: 1 (100%)
  Codes:
    200: 1
```

If you change `- header: ETag` to `- header: etag` the script executes both requests correctly.  I would expect the initial script to either work or throw an error.

```
All virtual users finished
Summary report @ 13:01:25(-0600) 2020-05-04
  Scenarios launched:  1
  Scenarios completed: 1
  Requests completed:  2
  Mean response/sec: 2.11
  Response time (msec):
    min: 116.5
    max: 558.6
    median: 337.6
    p95: 558.6
    p99: 558.6
  Scenario counts:
    0: 1 (100%)
  Codes:
    200: 1
    304: 1
```

Issue #802, opened 2020-02-14
Replacing Request.js with Got


hassy:
As you probably know [`Request.js`](https://github.com/request/request#deprecated) has been officially deprecated since Feb 11, 2020. Request.js is the HTTP library that Artillery uses under the hood.

Work on moving to [Got.js](https://github.com/sindresorhus/got) is already underway and is expected to complete soon.

## Why Got

* Lots of features, see the [feature comparison table](https://github.com/sindresorhus/got#comparison) for details. HTTP/2 support is especially promising.
* Actively developed & maintained.
* Migration looks to be quite straightforward.

## When

~~Soon™️. A preview release should be available in the next few days.~~

**Update 10/05** `master` is now fully-migrated to Got.js with full-backwards compatibility. The plan is to increase test coverage before marking the work as done.

hassy:
New version of Artillery's HTTP engine using Got is now on `master`.

enudler:
hi @hassy 
did you check the performance of this version with got vs previous version of request?
I could get up to 400 rps using pool of 250

using request.js I could hit 1200 rps.

hassy:
@enudler thanks for the info! I haven't run any benchmarks or profiled the code yet, but it's definitely something we'll look at before releasing the new engine. (Request.js will also still be available as a "legacy" engine for any users relying on Request.js-specific behavior.)

How complex is the scenario that you tested? i.e. are you using `capture`, custom hooks etc, or was it a simple request-in-a-loop type of scenario?

enudler:
thanks, @hassy 
it was simple get request
I will try to provide an example of the test and compare the vesrion

had also another issue when I received gzip responses
had to use decompress=false in got.js options.

enudler:
**test file:**
```
{
  "config": {
    "http": {
      "pool": 250
    },
    "target": "http://localhost:3001/webhooks",
    "phases": [{
      "duration": 60,
      "arrivalRate": 1000
    }]
  },
  "scenarios": [
    {
      "name": "Scenario 1",
      "weight": 100,
      "flow": [
        {
          "get": {
            "url": "http://localhost:3001/webhooks"
          }
        }
      ]
    }
  ]
}
```


**master:**

All virtual users finished

Summary report @ 14:50:38(+0300) 2020-03-31
Total time: 2 minutes, 18 seconds
  http.request_rate: ....................... 435/sec avg
  http.response_rate: ...................... 435/sec avg
[...]

PR #839, opened 2020-05-11
Feature/v1.6.x support case sensitive headers

jmclean-cnexus:
Allows for case sensitive headers, so long as config has below:
`caseSensitive: true`

If the above is not configured, default will be set to false, and prior functionality will persist

### DIFF
--- a/core/lib/engine_http.js
+++ b/core/lib/engine_http.js
@@ -31,6 +31,10 @@ const DEFAULT_AGENT_OPTIONS = {
 function HttpEngine(script) {
   this.config = script.config;
 
+  if(this.config.caseSensitive === 'undefined') {
+    this.config.caseSensitive = false;
+  }
+
   // If config.http.pool is set, create & reuse agents for all requests (with
   // max sockets set). That's what we're done here.
   // If config.http.pool is not set, we create new agents for each virtual user.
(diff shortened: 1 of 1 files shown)


jmclean-cnexus:
@hassy - please let me know your thoughts on this PR

hassy:
Thanks for the PR @jmclean-cnexus. I agree on the need to provide a way to maintain the casing of headers for servers that don't respect the HTTP spec. The complication we have to deal with is the move from Request.js to Got.js (#802). While Request respects the casing, Got.js seems to always lowercase request headers without a way to turn that off. Since the next major release of Artillery will use Got.js we'd need to find a way to make that work.

jmclean-cnexus:
@hassy agreed! For version 1.6.x, can we merge this. I would be happy to help out w/ the future release if needed. 1.6.x, I would imagine, would not incorporate Got.js.

jmclean-cnexus:
@hassy just following up on this. I can do a workaround for my local use case by just overwriting the source file w/ the one in this pull request...but i know other people were asking for this capability. BTW...any reason for doing Got.js v axios? just curious...i know there are 10000 ways to skin a cat, but curious what your motive is.

dandenton:
FYI this would fix https://github.com/artilleryio/artillery/issues/828 (just to link the two)

jmclean-cnexus:
@hassy I know you're a busy guy, but any/all comments are appreciated.

hassy:
Merged, thanks for the PR @jmclean-cnexus!

hassy:
and published in 1.6.1 now! thanks again @jmclean-cnexus!

BhavnaSoni1991:
@hassy , @bernardobridge Can anyone please comment if the fix for this issue is rolled back in latest releases as I am facing this issue in my test where the header name is getting changed to lower case and the tests are failing because of that?

PR #886, opened 2020-10-09
feat(http): add support for forwarding proxies

hassy:
Add support for using forwarding proxies via HTTP_PROXY and HTTPS_PROXY
environnment variables like Request.js [1]

1. https://github.com/request/request#controlling-proxy-behaviour-using-environment-variables


danjamesmay:
Thanks :+1:

PR #894, opened 2020-10-23
feat: allow plugins to run code before Artillery exits

hassy:
- Add support for adding extension points via the global `artillery` object
- Add a `beforeExit` extension point to run code before Artillery exits
- Add a way to suggest Artillery's exit code from plugins
- Remove current `config.ensure` code (to be replaced by a plugin)


hassy:
`artillery-plugin-ensure` using the new functionality to implement `config.ensure` support:

https://github.com/artilleryio/artillery-plugin-ensure

Issue #977, opened 2021-04-01
Cast phase config strings to numbers in order to support ENV vars


idpbond:
**The Issue**

When using ENV-based config for phases, for example: 
```yaml
config:
  phases:
    - duration: "{{ $processEnvironment.DURATION }}"
      arrivalRate: "{{ $processEnvironment.ARRIVAL_RATE }}"
      rampTo: "{{ $processEnvironment.RAMP_TO }}" # The troublemaker
```

The scenario does not run as expected (finishes early). In my anecdotal setup the `rampTo` bit is what introduces the issue.

**Why does this happen?**

The process environment variables which become the spec entry values are strings, and the various mathematical operations on these here:
https://github.com/artilleryio/artillery/blob/74baeac42a7bfb9ad9378ce7b9a01f4535eb1490/core/lib/phases.js#L85-L92

And here:
https://github.com/artilleryio/artillery/blob/74baeac42a7bfb9ad9378ce7b9a01f4535eb1490/core/lib/phases.js#L118

Are likely the culprit.

**The Solution**

PR introduces a conversion of defined phase config entries (based on a keys whitelist) expected to have numerical values.

hassy:
👍 PR merged, thank you!

idpbond:
Thanks for reporting @xyngfei 

@hassy I have an open PR that adds `maxVusers` as a cast value here: #1025

I believe, for now, that it's the only other phase config value in need of parsing, per [docs](https://artillery.io/docs/guides/guides/test-script-reference.html#Load-Phases)

idpbond:
@xyngfei 

1. Does this happen on artillery@1.7.2 ?
2. If so, can you provide a stripped down way (CLI or yml) to reproduce the issue?

xyngfei:
@idpbond 
Yes, I am using artillery@1.7.2, hope you can able to reproduce this error from your local environment

To run
export art_arrivalRate=1 art_rampto=20 art_sustain_duration=30 && ./node_modules/artillery/bin/artillery run --output report.json demo.yml

```
config:
  target: "https://example.com"
  plugins:
    expect: {}
  phases:
    - name: "ramp up"  
      duration: "{{ $processEnvironment.art_sustain_duration }}"
      rampTo: "{{ $processEnvironment.art_rampto }}"
      arrivalRate: "{{ $processEnvironment.art_arrivalRate }}"
scenarios:
  - name: "test demo site"
    flow:
      - post:
          url: "/api"
          capture:
[...]

PR #978, opened 2021-04-01
feat: cast string-valued phase configs

idpbond:
Allows for (eg: ENV-based) phase configuration values which are strings
by checking their definedness and casting them to numbers as necessary.
Previously, implementation was breaking under some (eg: rampTo) string
value configs. Addresses GH #977


hassy:
🎉 nice one, thank you @idpbond!

PR #997, opened 2021-04-19
Update CI scripts for building Docker image and publishing NPM package

dennmart:
This pull request contains a few updates to the CI scripts to help automate the Docker build and NPM publishing workflows.

By default, CircleCI doesn't execute workflows for any pushed tags unless specified, so pushing any Git tags never triggered the CI process. It's now updated to trigger those workflows when any tag starting with `v`, which is the current naming convention for releases.

I also updated the Docker build and NPM publish scripts that get executed with these CI workflows:

- The conditional check for `CIRCLE_BRANCH` was removed, since tagged pushes won't have this information.
- The command to check that the Docker image is working properly was changed, since `artillery quick` doesn't seem to valid command anymore. It's now going to use the `artillery dino` command, just to make sure we're not getting a non-zero exit code.
- In the Docker build script, if the pushed tag does not contain `-dev` as part of the string, we're assuming it's a stable release, so the built Docker image will get tagged as `latest`.
- Push all images (tagged build + `latest` if tagged) to Docker Hub.
- In the NPM publish script, if the pushed tag contains `-dev` as part of the string, we're assuming it's a development release, so the package will get published with the `dev` dist-tag. Otherwise, publish the package as the latest release.
- The Docker build will get tagged with the version number only according to the tag (e.g. if `CIRCLE_TAG` is `v1.7.0`, the image will be built and tagged as `artilleryio/artillery:1.7.0`).

I'm assuming that the Docker Hub and NPM credentials are set correctly on CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
       - run: npm run commitlint-last-commit
       - run: npm test
 
-  maybe_npm_publish:
+  npm_publish:
     working_directory: ~/repo
     docker:
       - image: circleci/node:12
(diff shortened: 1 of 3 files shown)


hassy:
🐳 🚀 Nice one!

PR #1015, opened 2021-05-03
fix(template): retain keys with periods

TwelveNights:
Preserves a situation where we have keys that could contain periods in it.

For example, `{ "hello.world": 1 }` would be templated as `{ "hello": { "world": 1 } }`, which could cause problems. One scenario I've encountered is when this effect mangles the request body for a request that expects keys that contain periods.


hassy:
🎉 thank you @TwelveNights!

PR #1017, opened 2021-05-04
test(template): add case for when the dotted key appears as actual path

TwelveNights:
Apologies for bringing in a PR so soon. This PR resolves an edge case when the path is something like `hello.world["hello.world"]`.

Beforehand, the template `{ hello: { world: { "hello.world": true } } }` would generate the object `{ hello: { world: { "hello.world": true } }, "hello.world": true }`. This fix ensures that the key isn't duplicated.

### DIFF
--- a/core/lib/engine_util.js
+++ b/core/lib/engine_util.js
@@ -259,7 +259,7 @@ function templateObjectOrArray(o, context) {
     }
 
     if (newPath.endsWith(key)) {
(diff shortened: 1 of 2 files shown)


hassy:
No worries, nice fix! Congratulations on becoming an Artillery contributor!

Drop me a note on h@artillery.io with an address we could send some swag to - if you'd like some that is! :)

PR #1041, opened 2021-05-21
fix: Expecting a falsy value always succeed

### LINKED ISSUE #1040: Matching false always succeed
I am trying to use socketio engine with artillery. My custom scenario requires acknowledge to be of type `{ status: false, data: ... }`. Even though socketio server replies with `{ status: true }`, no error is thrown. Same applies for all the falsy values of javascript.

```yaml
config:
  target: "http://localhost:3000/"
  socketio:
    transports: ["websocket"]
  phases:
    - duration: 1
      arrivalRate: 4
      maxVusers: 4

scenarios:
  - name: Test
    flow:
      - emit:
          channel: "custom_event"
          data: "{}"
          acknowledge:
            match:
              json: "$.0.status"
              value: false
```

mkadirtan:
Fixes #1040

### DIFF
--- a/lib/console-reporter.js
+++ b/lib/console-reporter.js
@@ -165,6 +165,7 @@ ConsoleReporter.prototype.printReport = function printReport(report, opts) {
       return name;
     }
 
+    report.rates = report.rates || {}
     const coreRates = Object.keys(report.rates).filter(name => name.startsWith('core.'));
     const engineRates = Object.keys(report.rates).filter(name => name.startsWith('engine.'));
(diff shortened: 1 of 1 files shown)


hassy:
Nice one @mkadirtan! Could you update the commit message to follow the existing guidelines - that will make the tests run on Circle too. (something like `fix: ability to match falsy values`) You can edit the commit message with `git commit --amend` in your branch, and then `git push --force` to it to update the PR.

mkadirtan:
There was a problem when I try to amend commit message. Here is the new PR #1045

PR #1045, opened 2021-05-24
fix: ability to match falsy values

### LINKED ISSUE #1040: Matching false always succeed

mkadirtan:
Fixes #1040

### DIFF
--- a/core/lib/engine_util.js
+++ b/core/lib/engine_util.js
@@ -420,7 +420,7 @@ function captureOrMatch(params, response, context, done) {
 
         let extractedValue = extractor(doc, template(expr, context), spec);
 
-        if (spec.value) {
+        if (spec.value !== undefined) {
           // this is a match spec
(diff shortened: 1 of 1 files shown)


hassy:
nice one, thank you @mkadirtan! 🎉

Issue #1065, opened 2021-06-08
Hooks are not available in socketio engine


DominicWatson:
I discovered that I could not use `beforeRequest`, `beforeScenario` et al in scenarios that used the SocketIo engine. Having these available is useful for various customisations, including setting cookies to be sent with requests.

shubham-studypad:
Hi, I just wanted to know if there is any update on this?

PR #1067, opened 2021-06-08
#1065 Add ability to use hooks with socketio scenarios

DominicWatson:
Being able to make use of hooks in socketio scenarios. Code stolen from the http engine and pasted here. I have tested locally and found that at least the `beforeScenario` hook is working just fine. 

Happy to add tests and refine if you're happy with conceptually doing this - but not a nodejs dev so lots to get up to speed with.


DominicWatson:
Can see tests failing, I'll look into this and will help with getting up to speed.

Issue #842, opened 2020-05-28
Run `beforeScenario` only once before a scenario


rvarbanov:
I thought that I can use the `beforeScenario` to call my auth service to get an access token and use that for each call to the service I'm load testing. However, my `auth` function is called before each endpoint call. That is suboptimal because I don't want load test my auth service at the same time I load test all my other services.

Code example:
```yml
scenarios:
  - name: "Endpoint returns 200"
    beforeScenario: "auth"
    flow:
      - get:
          url: "/api/v1/endpoint"
          expect:
            - statusCode: 200
            - contentType: json
```

AnkitGuhe:
@rvarbanov Wondering if you found the solution?

joberdick-rcd:
Is there really not a way to do a one time call that excludes being ran by all virtual users?

hassy:
`beforeScenario` is indeed meant to be run before each scenario execution (by each virtual user).

There is a way to run a call just once via a `before` section at the top level of the test definition. The contents of the section can be an arbitrary scenario, but any variables captured by that scenario will be available to all virtual users.

Quick example:

```yaml
config:
  target: https://artillery.io
  phases:
    - duration: 10
      arrivalRate: 1

before:
  flow:
    - log: "running a request before virtual users run"
    # Grab a random URL from the homepage:
    - get:
        url: "/"
        capture:
          - selector: "a"
            attr: "href"
            index: "random"
            as: "url"

scenarios:
  # All virtual users will make requests to the same URL
  - flow:
      - log: "URL is {{ url }}"
      - get:
          url: "{{ url }}"
```

savitanagaraj:
> `beforeScenario` is indeed meant to be run before each scenario execution (by each virtual user).
> 
> There is a way to run a call just once via a `before` section at the top level of the test definition. The contents of the section can be an arbitrary scenario, but any variables captured by that scenario will be available to all virtual users.
> 
> Quick example:
> 
> ```yaml
> config:
>   target: https://artillery.io
>   phases:
>     - duration: 10
>       arrivalRate: 1
> 
> before:
>   flow:
>     - log: "running a request before virtual users run"
>     # Grab a random URL from the homepage:
>     - get:
>         url: "/"
>         capture:
>           - selector: "a"
>             attr: "href"
>             index: "random"
>             as: "url"
> 
> scenarios:
>   # All virtual users will make requests to the same URL
>   - flow:
>       - log: "URL is {{ url }}"
>       - get:
[...]

Issue #751, opened 2019-10-02
after method


abarriel:
Hey ! I need to drop my database after the scenario is completed. there is anyway to achieve this ?
I know there is a `before` but it's not documented at all, I am wondering if it's the same for `after`
thank you

PR #1084, opened 2021-06-20
feat: allow running an after test flow

semiaddict:
There is already a (seemingly undocumented) way to run a flow before the actual test scenarios using the "[before](https://github.com/artilleryio/artillery/issues/842#issuecomment-765496311)" hook, which was added in https://github.com/artilleryio/artillery/commit/37bbe3b9319bf60063db32dbed369f57dd8d102f.

This MR addresses a similar need that seems to be requested by others as well (see https://github.com/artilleryio/artillery/issues/842#issuecomment-803206740 and https://github.com/artilleryio/artillery/issues/751) by adding a way to run a flow after the test scenarios have finished running. This can, for example, be useful to clean up test-only related data,

### DIFF
--- a/core/lib/runner.js
+++ b/core/lib/runner.js
@@ -158,7 +158,7 @@ async function runner(script, payload, options, callback) {
   //
   let contextVars;
   if (script.before) {
-    contextVars = await handleBeforeRequests(script, runnableScript, runnerEngines, ee);
+    contextVars = await handleScriptHook('before', runnableScript, runnerEngines, ee, contextVars);
   }
 
   //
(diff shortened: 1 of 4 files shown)


core/lib/runner.js
```diff
@@ -503,21 +512,21 @@ function $randomString(length) {
...
+function handleScriptHook(hook, script, engines, hookEvents, contextVars) {
   let ee = new EventEmitter();
   return new Promise(function(resolve, reject){
     ee.on('request', function() {
-      testEvents.emit('beforeTestRequest');
+      hookEvents.emit(`${hook}ScriptHookRequest`);
     });
     ee.on('error', function(error) {
-      testEvents.emit('beforeTestError', error);
+      hookEvents.emit(`${hook}ScriptHookError`, error);
```
flentini, line 522: I am not sure if we should change the names of these events as existing custom engines may rely on them

semiaddict: True. I just reverted those.

flentini:
Thank you for the PR @semiaddict !

Issue #1099, opened 2021-06-25
engine_http does not throw error if custom js function does not exist


elton-okawa:
Greetings everyone 👋 

I'm using `artillery v1.7.3` and noticed that `engine_http` does not throw an error if custom function does not exist [engine_http (195-209)](https://github.com/artilleryio/artillery/blob/7188c8134a2f8088284b8a8a2309a49e70b02d86/core/lib/engine_http.js#L195-L209)
```
if (requestSpec.function) {
  return function(context, callback) {
    let processFunc = self.config.processor[requestSpec.function];
    if (processFunc) {
      return processFunc(context, ee, function(hookErr) {
        if (hookErr) {
          ee.emit('error', hookErr.code || hookErr.message);
        }
        return callback(hookErr, context);
      });
    } else {
      return process.nextTick(function () { callback(null, context); });
    }
  };
}
```

I think that it'd be better to give an error right there instead of silently ignoring it, that way it'd help us a lot to debug an unusual behavior.

After discussion and approval, I'd like to try to change it.

hassy:
hi @elton-okawa 👋 great suggestion! and thank you for offering your time to implement it too!

the easiest way to implement this would be to use the [metrics API](https://artillery.io/docs/guides/guides/extending.html#Custom-metrics-API) and emit a counter for each of those errors. That would make the issue visible to the end user. Additional information (function name, stack trace etc) can be included in [`debug`](https://github.com/artilleryio/artillery/blob/master/core/lib/engine_http.js#L11) output if needed.

elton-okawa:
Solved by #1102

PR #1102, opened 2021-06-27
Fix/count error if function does not exist

elton-okawa:
Resolves issue: #1099

I've added tests files into folder `test/core/http` for better organization.
In test files, I used `lib/runner-sp` instead of `core/lib/runner`  like other core tests because of the processors load logic.

I also used v1.6.x as the origin branch, will it be merged into v2 in the future?


flentini:
thanks for the fix and for increasing the test coverage @elton-okawa ! I ported your changes to v2 as well

elton-okawa:
@flentini You're welcome! Also thank you for porting it to v2! 😄 

I'll close the referenced issue.

PR #1107, opened 2021-06-30
feat(run): add solo mode

hassy:
This adds the ability to run a script in "solo" mode, i.e. run a script with just one virtual user.

This is handy for debugging and testing your test scripts, and also for running functional tests. At the moment, you have to include a separate environment definition, with a phase definition that creates exactly one user.

This also means that it's now possible to run just a scenario on its own (no `config` required), with something like:

```sh
artillery run --target http://my-api.acme-corp.dev --solo scenario1.yml
```

### DIFF
--- a/lib/cmds/run.js
+++ b/lib/cmds/run.js
@@ -189,7 +189,7 @@ RunCommand.flags = {
   }),
   quiet: flags.boolean({
     char: 'q',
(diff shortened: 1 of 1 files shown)


flentini:
nice one! :ship:

PR #1112, opened 2021-07-02
Feat/socketio engine updates

flentini:
Updates socketio engine to use `socket.io-client` v4.1.2.
This is a porting to artillery core of https://github.com/ptejada/artillery-engine-socketio-v3.
 
`emit` must now be described as an array of items. This allows to pass an arbitrary number of arguments to the socketio emitter.

before:
```yaml
flow:
  - emit:
      channel: "echo"
      data: "Hello from Artillery"
``` 
after:
```yaml
flow:
  - emit:
    - "echo"
    - "Hello from Artillery"
```
multiple arguments can be passed to the socketio emit function: 
```yaml
flow:
  - emit:
    - "echo"
    - "Hello"
    - "from"
    - "Artillery"
```
Breaking changes:
options `acknowledge` and `response` must now live alongside the `emit` property, not under it.

before:
```yaml
flow:
  - emit:
      channel: "echo"
      data: "Hello from Artillery"
      response:
        channel: "echoResponse"
        data: "Hello from Artillery"
```
after:
```yaml
flow:
  - emit: ["echo", "Hello from Artillery"]
    response:
      channel: "echoResponse"
      data: "Hello from Artillery"
```


hassy:
Boom! 💥

PR #1129, opened 2021-07-16
fix: Update index.html.ejs

mersahin:
Wikipedia-Url #hash fixed


flentini:
thank you @mersahin !

PR #1203, opened 2021-11-09
feat: v2 reports

flentini:
this updates the reports to be generated using v2 metrics, using chartjs instead of metricgraphics


hassy:
:shipit:

PR #1219, opened 2021-11-11
fix(core): run before/after hook once

### LINKED ISSUE #1200: Before and After hooks are executed per worker in V2
The documented behaviour suggests that the before/after hooks are only executed once per test script. In V2 this seems to only be true if workers are limited to 1. Otherwise the setup and teardown routines are called multiple times (up to WORKERS count) which might lead to unexpected results.

Ideally this would be constrained to run before the scenario execution is forked and after all workers have completed or at least should be pointed out in the documentation.

flentini:
fixes #1200

### DIFF
--- a/core/lib/runner.js
+++ b/core/lib/runner.js
@@ -23,7 +23,7 @@ const wl = require('./weighted-pick');
 const Engines = {
   http: {},
   ws: {},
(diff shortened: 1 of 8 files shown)


hassy:
Nice 🚀

PR #1227, opened 2021-11-15
feat(reports): add custom counters charts and summary

flentini:
- adds a summary table for custom counters
- adds a custom counters at intervals line chart


hassy:
:shipit:

Issue #1254, opened 2021-12-06
config processor not working on pipeline


UnknownNS:
I am able to run the yaml file on local machine but as soon as I put in a pipeline it is giving me this error
WARNING: custom function getToken could not be found

file which uses it is like this
config:
    target: "https://xxx.xxx.com"
    processor: "./tokenFunction.js"
    http:
        maxSockets: 5
        timeout: 180
    phases:
        - name: Warm up phase
          duration: 300
          arrivalRate: 1
        - name: "22 requests/sec" 
          duration: 1200
          arrivalCount: 1200
    payload:
      - path: "./data.csv"
        fields:
          - "id"
before:
    flow:
        - log: "1. Get An Auth Token"
        - post:
            url: "https://xxx.xxx.com"
            beforeRequest: "getToken"

hassy:
hi @UnknownNS - can you double check that `tokenFunction.js` gets copied properly as well as the YAML file when running in the pipeline?

UnknownNS:
Ok the problem is on my local machine I was running the artillery version 1.7.9 and on the pipeline the artillery version is 2.0.0-4. As soon as I have updated the local machine version I have got the same error of  custom function getToken could not be found

flentini:
Hi @UnknownNS  - thank you for reporting this, I've been able to reproduce it on `v2.0.0-6`

PR #1259, opened 2021-12-10
fix: load script processor in runner

flentini:
so it's loaded once for both workers and main thread's hooks


hassy:
should fix https://github.com/artilleryio/artillery/issues/1254

lgtm, but worth having a test for it?

hassy:
🥳

PR #1264, opened 2021-12-15
fix: clone payload before passing it to before or after hooks running in the main thread

### LINKED ISSUE #1263: Issue when using csv payload in latest version
image
When try to load a payload from csv the console throw this error

Node version: v16.11.1

flentini:
fixes #1263

### DIFF
--- a/lib/launch-local.js
+++ b/lib/launch-local.js
@@ -13,6 +13,7 @@ const debug = require('debug')('core');
 
 const os = require('os');
 const p = require('util').promisify;
+const _ = require('lodash');
 
 const core = require('./dispatcher');
 const { handleScriptHook, prepareScript } = core.runnerFuncs;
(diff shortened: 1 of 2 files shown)


lib/launch-local.js
```diff
@@ -172,7 +173,7 @@ class Launcher {
       return {};
     }
 
-    const runnableScript = prepareScript(this.script, this.payload, this.opts);
+    const runnableScript = prepareScript(this.script, _.cloneDeep(this.payload), this.opts);
```
hassy, line 176: any reason not to clone `payload` inside `prepareScript`?

flentini: `prepareScript` is called by each worker, and by the time we are there the payload is already a copy so cloning it seemed redundant. it only needs to be cloned when the main thread uses it

Issue #683, opened 2019-04-21
Looping over an object's property


iamphi:
I would like to loop over using: `over: myObj.myArray`
My endpoint returns:
```
{
    "items": [ "one", "two", "three"]
}
```
and I capture with:
```
        capture:
          - json: "$"
            as: itemHolder
```
I cannot do 
```
    - loop:
      - log: "A item: {{$loopElement}}"
      over: itemHolder.items
```
My current work around is to capture:
```
          - json: "$.items"
            as: items
```
and do
```
    - loop:
      - log: "B item: {{$loopElement}}"
      over: items
```

my tests:
```config:
  target: "http://www.mocky.io/v2"
  phases:
    - duration: 1
      arrivalRate: 1
  variables:
    thingHolder:
      things: 
        - "A"
        - "B"
        - "C"
scenarios:
  - flow:
    - log: "thingHolder {{thingHolder}}"
    - log: "things {{thingHolder.things}}"
    - loop:
      - log: "thing {{ $loopElement }}"
      over: thingHolder.things
    - get:
        url: "/5cbc8de5320000e20e80d8e2"
        capture:
          - json: "$"
            as: itemHolder
          - json: "$.items"
            as: items
    - log: "itemHolder: {{itemHolder}}"
    - log: "items: {{items}}"
    - loop:
      - log: "A item: {{$loopElement}}"
      over: itemHolder.items
    - loop:
      - log: "B item: {{$loopElement}}"
      over: items```

hassy:
closed in #1273

PR #1273, opened 2021-12-23
feat: allow looping over object paths

OdysseyCShankland:
Original issue link: #683

This allows complex response objects to be used directly as loop over targets without the need to restructure them or otherwise duplicate values as top level context.vars properties. My use case merges data from multiple scenario steps to be used in a later loop, and allowing this greatly simplifies the logic.


hassy:
great PR, thank you @OdysseyCShankland!

PR #1283, opened 2022-01-06
fix: processor and plugins with 'before' hook 

flentini:
fixes a regression bug introduced in `2.0.0-8` and reverted in `2.0.0-9` where plugins could not attach functions to the test script if `processor` and `before` were both present


hassy:
🙌

PR #1296, opened 2022-01-19
add tests for http engine

flentini:


test/core/unit/engine_http.test.js
```diff
@@ -112,26 +118,210 @@ test('HTTP virtual user', function (t) {
...
+  const engine = new HttpEngine(script);
+  const ee = new EventEmitter();
+  const runScenario = engine.createScenario(script.scenarios[0], ee);
+
+  runScenario({ vars: {} }, function userDone(err, context) {
+    if (err) {
+      t.fail();
+    }
+
+    t.equal(context._successCount, 1, 'it should handle compressed responses');
```
hassy, line 271: * would it make sense to roll the check from the previous test (whether `accept-encoding` header is set) into here to reduce the amount of test code?
* `_successCount` does not seem to be used anywhere else in the codebase (even though it seems to be set by all engines) - would suggest not relying on it here and getting rid of it
* we're not actually testing that the response is decompressed by the client as expected here? (even checking `_successCount` does not tell us that we have the decompressed response)

test/core/unit/engine_http.test.js
```diff
@@ -141,11 +331,11 @@ test('url and uri parameters', function (t) {
     config: {
       target: 'http://localhost:8888',
       processor: {
-        rewriteUrl: function (req, context, ee, next) {
+        rewriteUrl: function (req, _context, _ee, next) {
```
hassy, line 334: should we add prefixing with `_`  to unused variables to lint rules?

test/core/unit/engine_http.test.js
```diff
@@ -247,21 +435,152 @@ test('hooks - afterResponse', (t) => {
...
+test('hooks - beforeScenario', (t) => {
+  const endpoint = '/products';
+  const script = {
+    config: {
+      target: 'http://localhost:8888',
+      processor: {
+        setEndpoint: function (context, ee, next) {
+          context.vars.endpoint = endpoint;
+
+          t.ok(
```
hassy, line 453: Is the goal here to check the signature of hook functions?

If so, should we have some assertions on other parameters too?

test/core/unit/engine_http.test.js
```diff
@@ -247,21 +435,152 @@ test('hooks - afterResponse', (t) => {
...
+
+test('hooks - afterScenario', (t) => {
+  const endpoint = '/products';
+  const productsCount = 123;
+  const script = {
+    config: {
+      target: 'http://localhost:8888',
+      processor: {
+        checkProductsCount: function (context, ee, next) {
+          t.ok(
```
hassy, line 511: as above - if the reason for this check is to check the interface of hook functions then we should check other args + maybe do it in one place to DRY?

test/core/unit/engine_http.test.js
```diff
@@ -247,21 +435,152 @@ test('hooks - afterResponse', (t) => {
...
+
+          t.equal(
+            context.vars.count,
+            productsCount,
+            'it can access variables set by the scenario'
+          );
+
+          t.ok(
+            context.vars.date === undefined,
+            'it cannot access variables set by other scenarios'
```
hassy, line 524: Nice idea - but it seems like the second scenario is never run, so the assertion can never fail?

test/core/test_capture.js
```diff
@@ -1,27 +1,71 @@
...
 
 let xmlCapture = null;
 try {
   xmlCapture = require('artillery-xml-capture');
 } catch (e) {}
 
+beforeEach(() => nock.cleanAll());
+
 test('Capture - headers', (t) => {
-  const fn = path.resolve(__dirname, './scripts/captures-header.json');
```
hassy, line 17: The file does not seem to be used anywhere else? should probably delete it altogether if so

test/core/test_capture.js
```diff
@@ -30,88 +74,178 @@ test('Capture - headers', (t) => {
...
-        t.assert(cond, 'There should be a 200 for every 201');
-        if (!cond) {
-          console.log(report);
-          console.log('200: %s; 201: %s', c200, c201);
-        }
-        ee.stop().then(() => {
-          t.end();
-        });
-      });
+test('Capture - selector', (t) => {
```
hassy, line 77: Nice 👍

test/core/test_capture.js
```diff
@@ -30,88 +74,178 @@ test('Capture - headers', (t) => {
   });
 });
 
-test('Capture - JSON', (t) => {
-  const fn = path.resolve(__dirname, './scripts/captures.json');
```
hassy, line 34: Should remove this and all other JSON scripts that have been inlined into tests if they're not used anywhere else.

test/core/test_capture.js
```diff
@@ -122,20 +256,96 @@ test('Capture before test - JSON', (t) => {
...
+              url: '/pets',
+              json: { name: '{{ name }}', species: '{{ species }}' },
+              capture: [
+                {
+                  json: '{{{ jsonPathExpr }}}',
+                  as: 'pet[id]'
+                },
+                {
+                  json: '$.doesnotexist',
+                  transform: 'this.doesnotexist.toUpperCase()',
```
hassy, line 290: `transform` can be removed (here and everywhere else)

test/core/test_capture.js
```diff
@@ -122,20 +256,96 @@ test('Capture before test - JSON', (t) => {
   });
 });
 
-test('Capture after test - JSON', (t) => {
-  const fn = path.resolve(__dirname, './scripts/after_test.json');
-  const script = require(fn);
-  runner(script).then(function (ee) {
+test('Capture and save to attribute of an Object in context.vars - JSON', (t) => {
```
hassy, line 259: Might be staring at right at it and not seeing it but what's the difference between this test and the previous one? :D

flentini: the first test captures a variable (`id`), the second captures a variable (`id`) as an object (`pet`) property

hassy: so the second one makes the first one unnecessary then?

PR #1315, opened 2022-01-29
chore: update chalk package to ^5.0.0

medanat:
Updates chalk package to latest fixing one vulnerability related to the `ansi-regex` package


medanat:
Based on `chalk` usage: https://github.com/artilleryio/artillery/search?q=chalk the API hasn't changed

PR #1372, opened 2022-03-14
fix: keep original load phase name when replacing arrivalCounts with pauses

### LINKED ISSUE #1313: Using arrivalCount will mess with phaseStarted and phaseCompleted events
I think with #1288 a race condition was introduced because only the first worker to start/complete a phase at the given index will be allowed to emit phaseStarted/phaseCompleted events to the reporter. With `artillery@2.0.0-11` that's causing phases showing up as "unnamed" in the console output at times.

flentini:
fixes #1313


hassy (approved):

hassy:
Nice 🥳

PR #1391, opened 2022-03-30
chore: migrate all tests to node-tap

chringwer:
Using two different testing libraries at the same time turns out to be problematic as both attempt to set the process exitCode. This race condition might cause test errors to be overshadowed by succesful tests run by the other library. Currently, there are a number of tests failing but still the build passes successfully. Choosing node-tap here over tape due to a lack of beforeEach/beforeAll hooks in tape that were used in some test files.


chringwer:
@hassy All 4 failing tests are also failing on `master`. I did not do any further investigation why that is, but at least the errors are now failing the build.

hassy:
wow, amazing, thank you @chringwer! 

happy to take over from here and fix up the failing test cases. If you could sign the CLA we can merge the changes in!

chringwer:
@hassy CLA is now signed. 👍

hassy:
merged, thank you @chringwer!

hassy:
tests now fixed in https://github.com/artilleryio/artillery/pull/1426! thanks again @chringwer 🥳

PR #1395, opened 2022-04-02
dockerfile: Run npm link

### LINKED ISSUE #1394: Latest Docker container doesn't include `artillery` command
The latest Docker container does not include the `artillery` command in the `/bin` directory. The only command available is `run`. `run` appears to be the `artillery` command, but does not match the usage documentation.

In the container, `run` must be used as `artillery`. e.g. `run run /path/to/script ...`

Latest container:

https://hub.docker.com/layers/artillery/artilleryio/artillery/2.0.0-14/images/sha256-8a1e0a395dbe0380b52b1632e73208c3fb0bc4d77eda262efdf5221bee5e8210

nzapo:
The currently built container image works fine if you want to
invoke artillery through a single docker command. But if you want to run
multiple artillery commands in the same container instance, or if you
want to use Artillery's container image as a base layer, then you ran
into issues such #1394.

The issue us that the executable in the path is named `run` instead of
`artillery`. To address this while maintain compatibility for any users
depending on the existing executable we can run an `npm install` which
will setup artillery as if it had been installed globally.

Fixes #1394.


YassinEldeeb (approved):
LGTM!

PR #1413, opened 2022-04-19
feat: Improving dev workflow

YassinEldeeb:
## What does this PR does?

1. Move the get in touch section to the end of the `CONTRIBUTING.md` file cause It's not that relevant to what the user has clicked.
2. Add Prerequisites & Setup guide for contributors.
3. Rename `bin/run` to `bin/run.js` to get the editor's language features and to be detected by `nodemon` watcher when developing.
4. end line in prettier is set to `CR` by default which is supposed to be the end line of old macs(pre-OSX), so better to set it to auto to pick up `CRLF` on Windows and `LF` on Linux/Mac and stop showing these annoying warnings on every line about line endings on windows.
5. Add start and dev scripts and use `nodemon` to watch file changes and run the CLI accordingly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,13 +1,5 @@
 # Artillery.io Contributors Guide
 
-## Need to get in touch?
-
-All project discussions should happen in the [issue tracker](https://github.com/artilleryio/artillery/issues) or on [Gitter](https://gitter.im/artilleryio/artillery).
-However if you are a first-time contributor and want some help getting started,
-feel free to get in touch over email:
-
-* Hassy Veldstra - [h@artillery.io](mailto:h@artillery.io?subject=Artillery Contribution Help)
-
 ## Guide for Contributions
 
 * We use the popular Fork+Pull model (more info here: [https://help.github.com/articles/using-pull-requests/](https://help.github.com/articles/using-pull-requests/)]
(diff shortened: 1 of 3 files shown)


YassinEldeeb:
@hassy can you review my PR, please?

hassy:
Thanks for the PR @YassinEldeeb!

Not sure about the prettier fix - we don't want a mix of line break characters in the repo from people committing from different operating systems. It should be possible to configure any decent editor to use a particular new line character regardless of the OS.

Not sure about how useful `nodemon` is in the context of a CLI.

Renaming `run` to `run.js` breaks the Dockerfile. It might also affect `oclif pack` - would need to check that.

YassinEldeeb:
> we don't want a mix of line break characters in the repo


I think git handles line endings conflicts on checkout brilliantly, [from their docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#example), though if there's a small percentage this would cause conflicts then I'm with you, we should probably ignore it.

IMG_20220421_162410

Yes, you're right in VSC for example I can set my line breaks to LF but I thought that it would be a problem with people using other editors that doesn't support this feature

hassy:
We don't need `nodemon` for the Artillery CLI, because it's a task-oriented CLI that runs one job to completion, rather than a long-running process like an HTTP server that sits there and waits for requests. There's no scenario in which I'd want the CLI restarted as the code changes - it will just exit immediately.

PR #1415, opened 2022-04-19
feat: switch package manager to pnpm(3x faster)

YassinEldeeb:
## What does this PR does?
This PR switches the default package manager `npm` with a faster, more modern alternative and has roughly the same API, so migrating wouldn't be a huge deal cause the commands are pretty similar.

### Performance Benchmark
These are speeds difference on my machine which would be very similar to what you'll get(pnpm is way faster):

```sh
pnpm: 11.7s
npm: 37.08s
```


YassinEldeeb:
I missed up this branch completely, so I'm gonna close this PR and follow up by another one.

PR #1416, opened 2022-04-19
chore: switch package manager to pnpm(3x faster)

YassinEldeeb:
## What does this PR does?
This PR switches the default package manager `npm` with a faster, more modern alternative and has roughly the same API, so migrating wouldn't be a huge deal cause the commands are pretty similar.

The API is very similar to that most people make `npm` as an alias for `pnpm` ([Source](https://twitter.com/jaredpalmer/status/1506625876691243013))

npm as alias for pnpm

### Performance Benchmark
These are speeds difference on my machine which would be very similar to what you'll get(pnpm is way faster):

```sh
pnpm: 11.7s
npm: 37.08s
```


hassy:
Thanks for the PR @YassinEldeeb. We're going to stick to npm as the official package manager. Individual contributors can use whatever package manager they prefer when developing locally.

PR #1422, opened 2022-04-23
feat: allow solo capture step with ws engine

GeekyDeaks:
It does not seem possible to `capture` a message from the server except as a response to a send.  This PR makes an alias of the `send` step called `wait` purely for readability purposes and also checks for a `payload` before sending anything, allowing a solo `capture` step to be defined.

The current ws engine also has an undocumented `match` feature, but you cannot use it without at least an empty capture.  This PR checks for either `capture` or `match` being defined rather than just `capture`.

e.g.

```
    flow:
      - send:
          payload:
            type: handshake
          capture:
            json: "$.user_id"
            as: "user_id"
      - think: 2
      - send:
          payload:
            type: text
            text: "unittest-20220420"
          match:
            json: "$.type"
            value: typing
      - wait:
          match: 
            json: "$.text"
            value: what is your name?
      - think: 2
      - send:
          payload:
            type: text
            text: "{{ naughtyString }}"
          match:
            json: "$.type"
            value: typing
      - wait:
          match: 
            json: "$.text"
            value: please enter yes
```

### DIFF
--- a/core/lib/engine_ws.js
+++ b/core/lib/engine_ws.js
@@ -64,9 +64,8 @@ function getMessageHandler(context, params, ee, callback) {
 
         const { captures = {}, matches = {} } = result;
 
-        debug('captures and matches:');
-        debug(matches);
-        debug(captures);
+        debug('matches: ', matches);
+        debug('captures: ', captures);
 
         // match and capture are strict by default:
         const haveFailedMatches = _.some(result.matches, function (v) {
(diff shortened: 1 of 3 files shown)


GeekyDeaks:
just realised my `npm test` didn't actually complete as it could not find `bats` so I missed the hanging proxy test.  I'll investigate now as I can still run it directly

hassy:
This is neat! and great find of the undocumented `match` feature @GeekyDeaks 😄 

Would be great to have an automated test for this, but I do appreciate that the current test codebase is quite messy.

GeekyDeaks:
hi @hassy - I have been taking a look at the tests and think I can knock something up.  Give me a day or two to get my head around it so I don't make a complete mess! :)

GeekyDeaks:
hi @hassy, I made two simple self-contained tests.  I only did two as that was the minimum required to test the changes and I thought it sensible to check the approach was ok before adding any others if necessary.

hassy:
this is great, thank you @GeekyDeaks!

tobyspark:
This functionality is undocumented, and pretty core to my mind. I went to create a documentation PR, but I don’t see any way to do that. Have I missed something?

PR #1424, opened 2022-04-26
Fix metric aggregation for local runs

juanig1:
moved here https://github.com/artilleryio/artillery/pull/1425


juanig1:
moved here https://github.com/artilleryio/artillery/pull/1425

PR #1430, opened 2022-04-30
feat: add timeout to ws engine

GeekyDeaks:
When capturing or matching responses with the `ws` engine, if the response doesn't arrive due to a server failure (not uncommon during a load test), the step will not complete which also prevents the load test from finishing

The PR adds a `config.ws.timeout` parameter to the `config` defined in seconds, similar to the `config.http.timeout` parameter

This PR also fixes some of the unit tests for the ws engine that incorrectly looked for errors when a `match` or `capture` failed, instead of checking `vusers.failed`


GeekyDeaks:
Hi @hassy - I am seeing the `Error: Callback was already called.` even on a clean clone of the master branch, so I don't think it's due to my changes, but I'm still investigating anyway.

hassy:
Looks like it was caused by https://github.com/artilleryio/artillery/pull/1412. taking a look

GeekyDeaks:
> Looks like it was caused by #1412. taking a look

I think you are right.  I just regressed it back to 1.5.2 and the tests are passing.  I'm also not seeing the same error occurring part way through a load test either.

GeekyDeaks:
hi @hassy - whilst working on this I realised some mistakes I made in my previous PR which this is also correcting.  I'm going to keep the fixes in separate commits (i.e. like the last one), but attach them to this PR if that is ok with you?

GeekyDeaks:
Hi @hassy, I think I have uncovered all the quirks I introduced in my previous PR now.  There are some minor things, but they are not affecting the engine functionally.  For example, I copied the timeout code from the http engine, but the error metric does not follow convention as it has a space, `response timeout` vs `response_timeout`.  Also, I refer to an `invalid_step` (https://github.com/artilleryio/artillery/pull/1430/files#diff-3260580d1e5c85e3971934bbc53a9b1495dec0f7272515f6d7a5e1e497f82c21R221) as it's in the `.step()` function, but the documentation refers to them as `actions` (https://www.artillery.io/docs/guides/guides/ws-reference#scenario-actions-and-configuration)

hassy:
Sounds good @GeekyDeaks, I'll take a closer look at the changes asap. 

Btw I'm reverting the `async` upgrade in https://github.com/artilleryio/artillery/pull/1432 for now. That test case fails consistently but I haven't had the chance to look into why.

hassy (approved):

PR #1461, opened 2022-05-28
feat: faster, smaller `deepEqual`

toastal:
The `deep-equal` dependency is both slower (100x) than `fast-deep-equal`
and ships a number of obsolete dependencies to polyfill browsers and Node
versions that this project doesn’t target (such as IE6 and Node.js
0.4.0).

To visualize the tree of dependencies `deep-equals` adds to your
project (requires JS enabled): https://npmgraph.js.org/?q=artillery

Shaving >6kb of minified dependiences and getting a speed up, moving to
`fast-deep-equal` seems like a win-win.


hassy:
very nice, thank you for the PR @toastal!

PR #1481, opened 2022-06-15
fix: html report styling

iamanoctopus:
Fixing up remaining styling gremlins.

### DIFF
--- a/lib/report/index.html.ejs
+++ b/lib/report/index.html.ejs
@@ -199,14 +199,10 @@
         <!-- json text output -->
         <div>
           <div class="hidden border-gray rounded" style="overflow: hidden;" id="json-output">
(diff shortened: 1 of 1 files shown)


iamanoctopus:
So, this one was just to tidy up some of the styling and bring things a little more closer to Control Panel. We'll also need to make some of the structural changes in Control Panel too (such as container width and breakpoints), but that's not relevant right now, Load Test view in general will be changing soon.

For now though, matched the latest look of Control Panel. We still have a problem with Chart.js responsiveness (this exists in Control Panel too), but it's not a huge deal breaker for the time being. As we'll change how Load Test views look it likely won't be an issue long term.

There we also some broken HTML tags, brushed up a little bit, but we can still simplify it more - well, depending on when we move to React or not. Also some Tailwind classes still aren't available here, so that move to React is much needed.

Below you can see the update version:

https://user-images.githubusercontent.com/606977/174283525-c1937ea8-1a35-4319-8992-9e1d7076e669.mov

Issue #1420, opened 2022-04-21
Setting cookies has no effect


MickL:
Version info: `2.0.0-14`

Setting cookies seem to have no effect. I dont receive any cookies in the incoming request on my Express server.

I tried using yaml:

```
scenarios:
  - name: "my scenario"
    flow:
      - post:
          url: /my-endpoint
          cookie:
            hello: 'world'
```

And also using beforeRequest:

```
export const beforeHandler: AsyncRequestInterceptor = async (
  request: ArtilleryRequest,
  context: ScenarioContext,
  ee: events.EventEmitter,
  next: Next
) => {
   request.cookie = {
     'hello': 'world'
   };

   return next();
}
```

MickL:
Seems to be a critical bug that kinda makes this whole project useless for now. @hassy Could you take a look at it maybe?

Issue #1035, opened 2021-05-19
You must set an arbitrary default cookie value in order to use the cookieJar to set a cookie


Shazwazza:
With the latest 1.7.x version the cookie jar (which is a [ToughCookie](https://github.com/salesforce/tough-cookie) instance) will not work unless it is specifically enabled. This means that if you have any JS that assigns a cookie to the request it will not work unless a previous response had a cookie, or your config has an explicit cookie defined. This is different than the prev 1.6.x version.

It's due to this code here: https://github.com/artilleryio/artillery/blob/v1.6.x/core/lib/engine_http.js#L262

To work around this you need to set a default cookie value, even if it's empty so the cookieJar is enabled. You can do this at top level of your config like:

```yml
config:
  defaults:
    cookie:
      MY_COOKIE: ""
```

or at request levels, etc...

If you do this, then the cookieJar is enabled which is this code: https://github.com/artilleryio/artillery/blob/v1.6.x/core/lib/engine_http.js#L653

There's another work around, which is to forcefully enable it in the beginRequest method like:

```js
if (!context._enableCookieJar) {
    context._enableCookieJar = true;
    requestParams.cookieJar = context._jar;
}
```

I think it may be clearer be able to explicitly enable this in config, something like

```yml
config:
  defaults:
    cookieJar: "always"
```

and perhaps you could have "auto" or even "never" (not sure why you'd want that, but maybe). Could be some other options like that but is a bit more clear that you just want to be able to dynamically set cookies in a request without having to declare an empty cookie.

Shazwazza:
(I also mostly wanted to report this in case other people are searching work arounds for this)

hassy:
Thank you for opening the issue @Shazwazza and for taking the time to investigate the cause! We can treat this as a regression - cookies should just work and be enabled by default.

PR #1492, opened 2022-06-28
Fix/no default cookies

juanig1:
When no default cookie is present, adding cookies for a particular scenario does not work

Closes [1035](https://github.com/artilleryio/artillery/issues/1035), [1420](https://github.com/artilleryio/artillery/issues/1420)


hassy:
LGTM! is this good to go @juanig1 or are you still working on it?

juanig1:
Found this [commit message](https://github.com/artilleryio/artillery/commit/1ef06ba9e322f96fbf0387243fb4f4ffb2a93104) warning about CPU usage.
Do you remember _how_ bad was it? Maybe adding a flag to disable the cookie jar may be a nice addition to this PR

hassy:
Ah, good find on that PR!

It looks like the real issue then is that when `set-cookie` header is present, the cookie jar is still not enabled. We should figure out why instead of enabling the cookie jar by default.

core/lib/engine_http.js
```diff
@@ -398,13 +398,14 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
         requestParams.headers = templatedHeaders;
 
         if (
-          typeof params.cookie === 'object' ||
+          typeof requestParams.cookie === 'object' ||
```
juanig1, line 401: requestparam includes the changes made by the processor

yaswanthpflegia (approved):

Issue #1536, opened 2022-07-23
worker error, id: 1 Error: Cookie failed to parse


MMaximus111:
Version info: 
```
docker image artilleryio/artillery:2.0.0-23

```

Running this command: 
```
docker run --rm -it -v ${PWD}:/scripts artilleryio/artillery run /scripts/front.yml
```

I expected to see this happen:

Load testing for my website.

Instead, this happened:

![image](https://user-images.githubusercontent.com/98421643/180619583-ba9d3d3f-6431-435f-92bc-507a4c17151a.png)

Files being used:

![image](https://user-images.githubusercontent.com/98421643/180619702-b48d7f61-b8c7-48dd-910c-6128ac75476a.png)

hassy:
hi @MMaximus111 👋 thank you for the bug report. Is your scenario (the part that's not in the screenshot) sending a cookie anywhere? (with `cookie` attribute on a request - https://www.artillery.io/docs/guides/guides/http-reference#cookies)

MMaximus111:
No, I didn't add a special attribute for cookies in my file.
 What is not included in the screenshot is similar.
 All methods differ only in the URL. 

![image](https://user-images.githubusercontent.com/98421643/180783856-45d1d082-75c5-4833-b642-2467f8085211.png)


The site itself returns some cookies, but for me they do not matter, I would even ignore them during load testing. In this scenario, unfortunately, because of them, something breaks.

If I can help you in any other way in finding a bug - let me know.

juanig1:
Hey, nice catch!

Managed to reproduce the error by targeting a server that returned a malformed Set-Cookie header like:
```
Set-Cookie: abcdef //or =abcdef
```
They should be ignored as per [RFC6265](https://datatracker.ietf.org/doc/html/rfc6265):
```
If the name-value-pair string lacks a %x3D ("=") character,
ignore the set-cookie-string entirely.
//...
If the name string is empty,
ignore the set-cookie-string entirely.
```

Going to go ahead and draft something that:
- ignore malformed cookies
- let user know, somehow
- continue with the scenario

Can you confirm this is what you are seeing on your end?
If not, please share your findings!

MMaximus111:
This is an example what i get in response about cookies. Maybe it can help you.
```
Response sent 0 bytes of Cookie data:
	Set-Cookie: 

Response sent 387 bytes of Cookie data:
[...]

PR #1540, opened 2022-07-25
Fix/handle malformed cookiestring response

juanig1:
See [this comment](https://github.com/artilleryio/artillery/issues/1536#issuecomment-1194431927)

tldr: handles malformed cookies received through a response Set-Cookie header
closes ART-574

### DIFF
--- a/core/lib/engine_http.js
+++ b/core/lib/engine_http.js
@@ -720,10 +720,10 @@ HttpEngine.prototype._handleResponse = function (
     if (rawCookies) {
       context._enableCookieJar = true;
       rawCookies.forEach(function (cookieString) {
-        if (tough.Cookie.parse(cookieString) !== undefined) {
+        try {
           context._jar.setCookieSync(cookieString, url);
-        } else {
-          console.log('Could not parse cookieString from response header, skipping it ' + cookieString);
+        } catch (err) {
+          ee.emit('err', 'Could not parse cookieString from response header, skipping it ' + cookieString);
         }
       });
     }
--- a/core/lib/engine_http.js
+++ b/core/lib/engine_http.js
@@ -720,7 +720,11 @@ HttpEngine.prototype._handleResponse = function (
     if (rawCookies) {
       context._enableCookieJar = true;
       rawCookies.forEach(function (cookieString) {
-        context._jar.setCookieSync(cookieString, url);
+        if (tough.Cookie.parse(cookieString) !== undefined) {
+          context._jar.setCookieSync(cookieString, url);
+        } else {
+          console.log('Could not parse cookieString from response header, skipping it ' + cookieString);
+        }
       });
     }
   }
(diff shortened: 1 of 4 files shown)


core/lib/engine_http.js
```diff
@@ -720,7 +720,11 @@ HttpEngine.prototype._handleResponse = function (
     if (rawCookies) {
       context._enableCookieJar = true;
       rawCookies.forEach(function (cookieString) {
-        context._jar.setCookieSync(cookieString, url);
+        if (tough.Cookie.parse(cookieString) !== undefined) {
+          context._jar.setCookieSync(cookieString, url);
+        } else {
+          console.log('Could not parse cookieString from response header, skipping it ' + cookieString);
```
juanig1, line 726: TODO: find best way to do this

hassy:
Is this a regression in v2.0.0-22, i.e. does it work with v2.0.0-21, or have we always had this issue?

core/lib/engine_http.js
hassy: Doesn't this way end up parsing every valid cookie twice? We should be able to catch the error thrown by `setCookieSync()` instead, right?

The error can be surfaced to the user by emitting an `error` event the way it's done elsewhere in the HTTP engine. Here for example: https://github.com/artilleryio/artillery/blob/74670ff2b3dcef341527615888e39348f5ac544c/core/lib/engine_http.js#L447

juanig1: Yeah, sound great, as failing is not remotely common.

Yeah saw the emit and also a log with a WARNING but wasn't sure.

I'll update both cases

core/lib/engine_http.js
```diff
@@ -720,7 +720,11 @@ HttpEngine.prototype._handleResponse = function (
     if (rawCookies) {
       context._enableCookieJar = true;
       rawCookies.forEach(function (cookieString) {
-        context._jar.setCookieSync(cookieString, url);
+        try {
+          context._jar.setCookieSync(cookieString, url);
+        } catch (err) {
+          ee.emit('error', 'Could not parse cookieString from response header, skipping it ' + cookieString);
```
hassy, line 726: The messages from these `error` events go right into the metric report, so it's best to keep it short. How about something like `cookie_parse_error_invalid_cookie`?

For debugging we can log the current message with `debug()` instead.

test/core/test_cookies.js
```diff
@@ -33,6 +34,23 @@ test('cookie jar http', function (t) {
   });
 });
 
+test('cookie jar invalid response', function (t) {
+  var script = require('./scripts/cookies_malformed_response.json');
+  runner(script).then(function (ee) {
+    ee.on('done', function (nr) {
+      const report = SSMS.legacyReport(nr).report();
+      t.ok(
+        report.codes[200] && report.codes[200] > 0,
```
hassy, line 43: We should check that the errors are actually reported in this test.

core/lib/engine_http.js
juanig1: Done, used uppercase for the error message for consistency.
One detail is that we can get N error messages for one call (as there may be N invalid Set-Cookie headers)

PR #1544, opened 2022-07-28
fix(http): prevent multiple callbacks when a capture/match fails

### LINKED ISSUE #1542: Artillery going in infinite loop with "Callback was already called" error
Hi @hassy :

After the latest update you made with the callback error, I reinstalled the latest version of artillery using the below command:
```npm install -g artillery@v2.0.0-22```

Ran the below command to execute artillery script and I am still getting the same error and the run is going in a infinite loop.

```artillery run my-file.yaml -e dev --quiet```

Below is the error I am still getting:

```
worker error, id: 1 Error: Callback was already called.
    at /usr/local/lib/node_modules/artillery/node_modules/async/dist/async.js:969:32
    at /usr/local/lib/node_modules/artillery/core/lib/engine_http.js:698:20
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

### PR DESCRIPTION
**Pull request by hassy**
Fixes #1542

### DIFF
--- a/core/lib/engine_http.js
+++ b/core/lib/engine_http.js
@@ -221,6 +221,7 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
   let f = function (context, callback) {
     // Previous request had an error of the kind that should stop current VU
     if (context.error) {
+      ee.emit('error', err.message || err.code);
       return callback(context.error, context);
     }
 
@@ -833,7 +834,6 @@ HttpEngine.prototype.compile = function compile(tasks, scenarioSpec, ee) {
 
     async.waterfall(steps, function scenarioWaterfallCb(err, context) {
       if (err) {
-        ee.emit('error', err.message);
         return callback(err, context);
       } else {
         return callback(null, context);
--- a/core/lib/engine_http.js
+++ b/core/lib/engine_http.js
@@ -219,6 +219,11 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
   }
 
   let f = function (context, callback) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by meetveera** [comment]
Hi @hassy - Good Morning. Thank you very much for fixing the problem. Now I am getting proper status codes for multiple redirects errors. 
But again for some APIs it is still going on infinite loop. I double checked my API by executing it via postman and it is working fine there. When I did a little more digging and commented the below line from the snippet from my local machine that all my APIs started to work absolutely fine. 
```.catch((gotErr) => {
            // TODO: Handle the error properly with run hooks
            ee.emit('error', gotErr.code || gotErr.message);
           // return callback(gotErr, context); line 702 from engine_http.js
 });```

Could you please take a look and see what exactly the problem is? I might be wrong but again commenting that line just worked for me and now all my apis are working fine via artillery run.

Thanks

**Comment by hassy** [comment]
@meetveera could you try the most recent release of Artillery to see if it fixes the issue.

**Comment by onematchfox** [comment]
> @meetveera could you try the most recent release of Artillery to see if it fixes the issue.

Hi @hassy,

I'm also running into this issue - in the latest version `2.0.0-23`. In my case it presents when looping over requests that are set to expect a specific status code (in one case a `403` and in another a `302`). E.g.

``` yaml
- loop:
    - get:
        url: '/foo?bar={{ $loopElement }}'
        followRedirect: false
        expect:
          - statusCode: 302
  over: myVariable
- loop:
    - get:
        url: '/{{ $loopElement }}'
        expect:
          - statusCode: 403
  over: myVariable
```

Let me know if I should open a separate issue for this.

PR #1551, opened 2022-08-01
Update README.md

stavares843:
Minor suggestion on README.md 🔨


hassy:
thank you @stavares843! we use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), so the commit message needs to be amended. Good to go otherwise.

stavares843:
https://github.com/artilleryio/artillery/pull/1555

Issue #1562, opened 2022-08-04
Artillery 2.0.0-22 Gzip Decoding + capture doesn't work


luneo7:
Version info: 
```
Artillery Core: 2.0.0-22
Artillery Pro:  not installed (https://artillery.io/product)

Node.js: v16.15.0
OS:      darwin
```

Running this command: 
```
artillery run -e env script.yml
```

I expected to see this happen:

*Gzip decoding happens and capture works*

Instead, this happened:

*Gzip decoding didn't happen and body was left with empty string*

```
load_test/running-tests/load_test/node_modules/artillery/core/lib/engine_http.js:613
                      new Error('Failed capture or match'),
                      ^

Error: Failed capture or match
    at load_test/running-tests/load_test/node_modules/artillery/core/lib/engine_http.js:613:23
    at load_test/running-tests/load_test/node_modules/async/dist/async.js:473:16
    at replenish (load_test/running-tests/load_test/node_modules/async/dist/async.js:1009:25)
    at load_test/running-tests/load_test/node_modules/async/dist/async.js:1019:9
    at eachLimit$1 (load_test/running-tests/load_test/node_modules/async/dist/async.js:3199:24)
    at Object.<anonymous> (load_test/running-tests/load_test/node_modules/async/dist/async.js:1049:16)
    at captured (load_test/running-tests/load_test/node_modules/artillery/core/lib/engine_http.js:578:21)
    at load_test/running-tests/load_test/node_modules/artillery/core/lib/engine_util.js:488:16
    at load_test/running-tests/load_test/node_modules/async/dist/async.js:473:16
    at replenish (load_test/load_test/restaurant-orderflow-staging/node_modules/async/dist/async.js:1009:25
```

Files being used:
script.yml
```
config:
  environments:
    env:
      target: "https://api-staging.blahblah.com"
      phases:
        - name: "Test 1"
          duration: 1
          arrivalRate: 1
        - pause: 60
        - name: "Break"
          duration: 1
          arrivalRate: 0
        - name: "2 times peak load"
          duration: 2
          arrivalRate: 2
        - name: "Break"
          duration: 1
          arrivalRate: 0
  payload:
    - path: "payloads/{{ $environment }}-payload.csv"
      fields:
        - "restaurant_id"
        - "email"
        - "password"
        - "etag"
        - "env"
      skipHeader: true
  processor: "./processor.js"
before:
  flow:
    - log: "DOING_LOGIN - env {{ env }}"
    - post:
[...]

PR #1563, opened 2022-08-04
fix(http): handle response decompression properly

luneo7:
Almost the same thing as https://github.com/artilleryio/artillery/pull/999

Since we are using `decompressResponse` and `on('data')` is added to it, `on('end')` should also be executed on the decompressResponse... but it is being executed on raw response causing a race condition, so when response end happens, response data didn't happen yet, and we end up with empty body string.

This PR makes the handle response use decompressResponse for everything instead of mixing raw response object and the decompress response object.

Fixes https://github.com/artilleryio/artillery/issues/1562

### DIFF
--- a/core/lib/engine_http.js
+++ b/core/lib/engine_http.js
@@ -713,10 +713,10 @@ HttpEngine.prototype._handleResponse = function (
   callback
 ) {
   const url = requestParams.url;
-  const decompressedRes = decompressResponse(res);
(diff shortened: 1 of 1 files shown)


douglasmelo (approved):

wcaquino (approved):

murillomarques (approved):
Nice work.

hassy:
nice one @luneo7, thanks for your PR!

PR #1568, opened 2022-08-17
fix: redirect chains with capture/afterResponse hooks

hassy:
This PR improves our handling of redirect chains.

We currently run response processors on every response, so e.g. a `capture` can fail for the first response but succeed for the second response, leading to multiple callbacks to `async.waterfall` being invoked.

This PR changes the logic to only run response processors on the last request in the chain.

It also refactors the code to get rid of `async.waterfall`.


hassy:
Related: https://github.com/artilleryio/artillery/pull/1544

Issue #1575, opened 2022-08-23
tough.CookieJar() preventing access to some sites with "Could not parse cookieString" error


theTestTube:
Hi! 
      
Firstly, thanks for your work on Artillery! Smart testing framework with simplicity in mind. 👋 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `artillery@2.0.0-23` for the project I'm working on.

Tests run fail with a summary message `errors.cookie_parse_error_invalid_cookie`. 

Running with `DEBUG="http"` I can see,

``` text
| 2022-08-23T18:22:09.276Z http Could not parse cookieString "GUEST_LANGUAGE_ID=es_ES; Domain=.example.net; Expires=Wed, 23-Aug-2023 18:22:09 GMT; Path=/; HttpOnly" from response header, skipping it
2022-08-23T18:22:09.323Z http request: {
  "url": "https://mydomain.example.net/",
  "method": "GET",
  "headers": {
    "user-agent": "Artillery (https://artillery.io)"
  }
```

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/artillery/core/lib/engine_http.js b/node_modules/artillery/core/lib/engine_http.js
index fb72c99..588fd55 100644
--- a/node_modules/artillery/core/lib/engine_http.js
+++ b/node_modules/artillery/core/lib/engine_http.js
@@ -776,7 +776,7 @@ HttpEngine.prototype.setInitialContext = function (initialContext) {
 
   initialContext._defaultStrictCapture = this.config.defaults.strictCapture;
 
-  initialContext._jar = new tough.CookieJar();
+  initialContext._jar = new tough.CookieJar(undefined, false);
   initialContext._enableCookieJar = false;
   // If a default cookie is set, we will use the jar straightaway:
   if (typeof this.config.defaults.cookie === 'object') {
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

hassy:
hi @theTestTube, thank you for the bug fix!

I'm not sure why that change solves the issue though 🤔 Looking at the `CookieJar` API, passing `undefined` as the first argument should make it default to using in-memory cookie storage, which it's doing already. Passing `false` as the second parameter *presumably* is ignored by the constructor as it expects an object, so the behavior should be the same.

Docs here: https://github.com/salesforce/tough-cookie#cookiejarstore-options

theTestTube:
[...]

PR #1581, opened 2022-09-01
feat: allow customization of cookie parsing behavior

hassy:
config.http.cookieJarOptions may be set to customize the behavior
of the underlying ToughCookie instance.

See available options in ToughCookie docs:

https://github.com/salesforce/tough-cookie#cookiejarstore-options


hassy:
Fixes #1575

PR #1620, opened 2022-10-14
fix: add needed flow prop to scripts

juanig1:
Current scripts do not work as they are missing a `flow` property under `scenario`

### DIFF
--- a/examples/browser-load-testing-playwright/browser-load-test.yml
+++ b/examples/browser-load-testing-playwright/browser-load-test.yml
@@ -10,3 +10,4 @@ scenarios:
   - name: "Dev account signup"
     engine: playwright
     flowFunction: "devAccountSignupFlow"
+    flow: []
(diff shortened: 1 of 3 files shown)


hassy:
Nice! would it make sense to update the versions in `package.json` as well? (this might be why the prop isn't there already - the example is pinned to an older version of Artillery)

juanig1:
Nice catch, seems `flow: []` is still needed tho.

Regarding `package.json`, want me to update all of the examples to point to `latest` of anything artillery related?
Or pinning to actual versions, but that may result in similar issues in the future.

hassy:
> Nice catch, seems flow: [] is still needed tho.

Hmm, how are you running the example? It works fine for me with:

```
$(npm bin)/artillery run browser-load-test.yml
```

Either way, let's update everything to latest versions for this example. 👍 I think pinning is preferable, as the example is guaranteed to continue working.

juanig1:
Excellent, will pin to current ones!

`$(npm bin)/artillery run browser-load-test.yml`
this works on `artillery/2.0.0-6` but not on `2.0.0-23`, so there's something else to dive into

Issue #1589, opened 2022-09-07
Ramp Up phase produces spike in VUs


tchambers97:
Version info: 
```
Artillery Core: 2.0.0-23
```

With these phases and scenarios defined in my config:
```
  phases:
    - duration: 60
      arrivalCount: 1
      name: Start Up
    - duration: 900
      arrivalRate: 1
      rampTo: 25
      name: Ramp Up

scenarios:
  - name: HomePage
    engine: playwright
    flowFunction: 'homePage'
    flow: []
    weight: 5
  - name: PlpPage
    engine: playwright
    flowFunction: 'plpPage'
    flow: []
    weight: 3
  - name: PdpPage
    engine: playwright
    flowFunction: 'pdpPage'
    flow: []
    weight: 2

```

I expected to see this happen:

Linear-ish ramp up from 1 to 25 users over 15 minutes

Instead, a large spike in users was created at the beginning of the ramp up phase, followed by timeout errors and most of the time the worker dies (some metrics and errors removed for brevity):

```
Running scenarios...
Phase started: Start Up (index: 0, duration: 60s) 11:15:52(-0400)

Phase completed: Start Up (index: 0, duration: 60s) 11:16:52(-0400)

Phase started: Ramp Up (index: 1, duration: 900s) 11:16:52(-0400)

--------------------------------------
Metrics for period to: 11:17:00(-0400) (width: 7.948s)
--------------------------------------

engine.browser.http_requests: .................................................. 1874
engine.browser.memory_used_mb:
  min: ......................................................................... 13.4
  max: ......................................................................... 21.3
  median: ...................................................................... 15.6
  p95: ......................................................................... 20.3
  p99: ......................................................................... 20.3
user.status_check_ok.: .......................... 8
user.status_check_ok.shop/mens: ................ 1
vusers.completed: .............................................................. 8
vusers.created: ................................................................ 71
vusers.created_by_name.HomePage: ............................................... 33
vusers.created_by_name.PdpPage: ................................................ 8
vusers.created_by_name.PlpPage: ................................................ 30
vusers.failed: ................................................................. 0
vusers.session_length:
  min: ......................................................................... 2419.9
  max: ......................................................................... 6329.4
  median: ...................................................................... 2671
  p95: ......................................................................... 6187.2
  p99: ......................................................................... 6187.2


--------------------------------------
Metrics for period to: 11:17:10(-0400) (width: 9.919s)
--------------------------------------

[...]

PR #1621, opened 2022-10-17
fix: rampTo behaviour when arrivalRate is less than amount of workers

juanig1:
See #1589.
Setting an arrival rate lower than the workers being user to run the test, resulted in `rampTo` property being distributed across all workers alongside an incorrect value for `arrivalRate`

This generated an unexpected distribution of requests:
The root cause for this is that having a `rampTo` and no `arrivalRate`, defaulted to `arrivalRate` being 1.
Example:
using `arrivalRate:1; rampTo:25` and running with 7 workers resulted in:

arrival rate distribution for each worker:
```
[
  1, 0, 0, 0, 0, 0, 0
]
```

rampTo distribution for each worker:
```
[
  4, 4, 4, 4, 3, 3, 3
]
```

When this got to the execution logic, arrival rate was being updated to 1 [here](https://github.com/artilleryio/artillery/blob/master/core/lib/phases.js) causing the initial load to be higher than expected, so effective arrival rates were actually:
```
[
  1, 1, 1, 1, 1, 1, 1 // 0s -> 1s
]
```

After the update we remove 1 as default  `arrivalRate`, avoiding the initial spike and still getting the most out of available threads.
Still worth mentioning that for smaller numbers setting `WORKERS=1` may be desired to have a more controlled behavior.

### DIFF
--- a/core/lib/phases.js
+++ b/core/lib/phases.js
@@ -94,7 +94,7 @@ function createPause(spec, ee) {
 
 function createRamp(spec, ee) {
   const duration = spec.duration || 1;
-  const arrivalRate = spec.arrivalRate || 1;
+  const arrivalRate = spec.arrivalRate || 0;
   const rampTo = spec.rampTo || arrivalRate + 1;
 
   const tick = 1000 / Math.max(arrivalRate, rampTo); // smallest tick
--- a/lib/dist.js
+++ b/lib/dist.js
@@ -30,7 +30,11 @@ function divideWork(script, numWorkers) {
       });
     }
 
-    if (phase.arrivalRate && phase.rampTo) {
+    if (phase.rampTo) {
+      // same behaviour as a single worker. Now we support scripts
+      // with rampTo and no arrivalRate
+      phase.arrivalRate = phase.arrivalRate || 0;
+
       let rates = distribute(phase.arrivalRate, numWorkers);
       let ramps = distribute(phase.rampTo, numWorkers);
       L.each(rates, function (Lr, i) {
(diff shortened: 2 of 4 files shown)


core/lib/phases.js
```diff
@@ -94,7 +94,7 @@ function createPause(spec, ee) {
 
 function createRamp(spec, ee) {
   const duration = spec.duration || 1;
-  const arrivalRate = spec.arrivalRate || 1;
+  const arrivalRate = spec.arrivalRate || 0;
```
juanig1, line 97:
```suggestion
```

test/core/unit/phases.test.js
```diff
@@ -68,6 +68,21 @@ test('pause', function (t) {
   phaser.run();
 });
 
+test('rampTo with no arrivalRate defaults to 0', function (t) {
```
hassy, line 71:
```suggestion
test('arrivalRate set to 0 stays at 0', function (t) {
```

test/core/unit/dist.test.js
```diff
@@ -2,6 +2,8 @@
 
 const tap = require('tap');
 const divideWork = require('../../../lib/dist');
+const isIdlePhase = require('../../../core/lib/is-idle-phase');
```
hassy, line 5:
```suggestion
```

Does not seem to be used.

lib/dist.js
```diff
@@ -30,7 +30,11 @@ function divideWork(script, numWorkers) {
       });
     }
 
-    if (phase.arrivalRate && phase.rampTo) {
+    if (phase.rampTo) {
+      // same behaviour as a single worker. Now we support scripts
+      // with rampTo and no arrivalRate
+      phase.arrivalRate = phase.arrivalRate || 0;
+
```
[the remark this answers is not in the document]
hassy: Should be OK!

hassy (approved):
Couple of minor suggestions. LGTM otherwise!

PR #1633, opened 2022-10-24
simpler rampup logic

juanig1:
Define a linear function f(t) = describing the 'ideal' VUs for a given time 
  - Encapsulate it in `arrivalProbability(step)` that returns the probability [0,1] that we need to spawn a new arrival
  - We set the interval between arrivals as the biggest one needed to reach `max(rampTo, rampFrom)` throughput
  - In each step we `roll` and compare that to the function
closes ART-664

### DIFF
--- a/core/lib/phases.js
+++ b/core/lib/phases.js
@@ -97,7 +97,7 @@ function createRamp(spec, ee) {
   const arrivalRate = spec.arrivalRate;
   const rampTo = spec.rampTo;
 
-  const tick = 1000; // 1s; match VUs arrival rate
+  const tick = 1000 / Math.max(arrivalRate, rampTo); // smallest tick
   const difference = rampTo - arrivalRate;
   const periods = duration * 1000 / tick;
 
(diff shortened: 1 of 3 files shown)


core/lib/phases.js
```diff
@@ -94,83 +94,46 @@ function createPause(spec, ee) {
...
+  function arrivalProbability(currentStep) {
+    // linear function ax + b
+    // normalized to 0 <= f(t) <= 1
+    // Anything under function value should be an arrival
+
+    let t = currentStep * tick / 1000;
+    return ((difference / duration) * t + arrivalRate) / Math.max(difference, arrivalRate);
+  };
+
+  let probabilities = Array.from({length: periods}, () => Math.random());
```
juanig1, line 113: Wasn't really sure we need a cryptographically secure generation here. Using `Math.random` should be more performant

test/testcases/command-run.bats
```diff
@@ -111,3 +111,21 @@
 
     [[ $status -eq 1 ]]
 }
+
+@test "Running ramp up script throughput does not depend on workers" {
```
juanig1, line 115:
```suggestion
@test "Ramp up script throughput does not depend on workers" {
```

test/testcases/command-run.bats
```diff
@@ -111,3 +111,21 @@
...
+    ./bin/run run -o multiple_workers.json test/scripts/ramp.json
+    WORKERS=1 ./bin/run run -o single_worker.json test/scripts/ramp.json
+
+    multiple_count=$(jq '.aggregate.counters."vusers.created"' multiple_workers.json)
+    single_count=$(jq '.aggregate.counters."vusers.created"' single_worker.json)
+
+    echo $multiple_count
+    echo $single_count
```
juanig1, line 123:
```suggestion
    echo $multiple_count
    echo $single_count
```

PR #1638, opened 2022-10-31
test: fix flaky ramp up test

juanig1:
Compare values to expected total and use 20% as threshold instead of comparing multiple workers vs single workers results.

We are currently running on [`Medium`](https://circleci.com/docs/configuration-reference/#docker-execution-environment) instances with two cores

### DIFF
--- a/test/testcases/command-run.bats
+++ b/test/testcases/command-run.bats
@@ -122,7 +122,12 @@
     rm multiple_workers.json
     rm single_worker.json
 
-    diff=$((single_count-multiple_count))
-    diff_abs=${diff#-}
-    [ $diff_abs -le 10 ]
+    expected=55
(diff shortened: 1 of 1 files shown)


hassy:
Nice! Two possible improvements to consider:

1. How many CPUs do we have in Circle? If it's not more than one, the test may not be exercising multi-thread logic.
2. Comparing the total number of arrivals would not catch issues with how those arrivals are distributed, e.g. this test would not have caught the issue with the previous implementation of `rampTo` where with certain values of arrivalRate & duration all VUs would arrive at the end of the phase.

wdyt?

hassy:
merging to unblock main (the current test is quite flaky)

let's handle setting worker numbers & checking arrivals in another PR

PR #1654, opened 2022-11-07
fix: test with no explicit phases do not create final report in distributed mode

juanig1:
Stat aggregation waits for all workers to emit the `stats`, but sleeping workers did not.
If this happened in the last phase of the test, there was a race condition where one possible scenario was the test ending before final report was printed.

Added emit to sleeping worker phases, refactored `isIdlePhase` so that sleeping workers are not taken into account as needed workers to build report.
  - Short tests are going to emit the final report properly as we get rid of the race condition where report aggregation and app shutdown are competing
  - Short tests are going to finish sooner (3-4s in my machine but its tied to number of workers)
Closes ART-642

### DIFF
--- a/core/lib/is-idle-phase.js
+++ b/core/lib/is-idle-phase.js
@@ -6,7 +6,8 @@ function isIdlePhase(phase) {
   return (
     (phase.arrivalRate === 0 && !phase.rampTo) ||
     phase.arrivalCount === 0 ||
-    phase.maxVusers === 0
+    phase.maxVusers === 0 ||
+    phase.pause > 0
   );
 }
 
--- a/core/lib/runner.js
+++ b/core/lib/runner.js
@@ -200,14 +200,11 @@ function run(script, ee, options, runState, contextVars) {
   });
   phaser.on('phaseStarted', function (spec) {
     ee.emit('phaseStarted', spec);
-    // if (isIdlePhase(spec)) {
-    //   ee.emit('stats', SSMS.empty());
-    // }
+    if (spec.pause > 0) {
+      ee.emit('stats', SSMS.empty());
+    }
   });
   phaser.on('phaseCompleted', function (spec) {
-    // if (isIdlePhase(spec)) {
-    //   ee.emit('stats', SSMS.empty());
-    // }
     ee.emit('phaseCompleted', spec);
   });
   phaser.on('done', function () {
--- a/core/lib/runner.js
+++ b/core/lib/runner.js
@@ -200,12 +200,15 @@ function run(script, ee, options, runState, contextVars) {
   });
   phaser.on('phaseStarted', function (spec) {
     ee.emit('phaseStarted', spec);
-    if (spec.pause > 0) {
+    if (isIdlePhase(spec)) {
       ee.emit('stats', SSMS.empty());
     }
   });
   phaser.on('phaseCompleted', function (spec) {
     ee.emit('phaseCompleted', spec);
+    if (isIdlePhase(spec)) {
+      ee.emit('stats', SSMS.empty());
+    }
   });
   phaser.on('done', function () {
     debug('All phases launched');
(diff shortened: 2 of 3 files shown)


hassy:
I wonder why the bit of commented out code that this change is replacing was commented out in the first place. 🤔 It looks like it would've resulted in the same behavior. `git blame` is not of much help unfortunately - but it looks like it was done as part of the change implementing distribution across worker threads:

```sh
git blame core/lib/runner.js -L 200,213
```

```
^7fb802d lib/runner.js      (hveldstra 2015-07-10 15:20:49 +0100 200)   });
33209c1e core/lib/runner.js (hveldstra 2022-01-06 15:05:12 +0000 201)   phaser.on('phaseStarted', function (spec) {
31382808 lib/runner.js      (hveldstra 2015-12-18 22:46:37 +0000 202)     ee.emit('phaseStarted', spec);
8beb82c4 core/lib/runner.js (hveldstra 2021-08-05 12:23:42 +0100 203)     // if (isIdlePhase(spec)) {
8beb82c4 core/lib/runner.js (hveldstra 2021-08-05 12:23:42 +0100 204)     //   ee.emit('stats', SSMS.empty());
8beb82c4 core/lib/runner.js (hveldstra 2021-08-05 12:23:42 +0100 205)     // }
31382808 lib/runner.js      (hveldstra 2015-12-18 22:46:37 +0000 206)   });
33209c1e core/lib/runner.js (hveldstra 2022-01-06 15:05:12 +0000 207)   phaser.on('phaseCompleted', function (spec) {
8beb82c4 core/lib/runner.js (hveldstra 2021-08-05 12:23:42 +0100 208)     // if (isIdlePhase(spec)) {
8beb82c4 core/lib/runner.js (hveldstra 2021-08-05 12:23:42 +0100 209)     //   ee.emit('stats', SSMS.empty());
8beb82c4 core/lib/runner.js (hveldstra 2021-08-05 12:23:42 +0100 210)     // }
31382808 lib/runner.js      (hveldstra 2015-12-18 22:46:37 +0000 211)     ee.emit('phaseCompleted', spec);
31382808 lib/runner.js      (hveldstra 2015-12-18 22:46:37 +0000 212)   });
33209c1e core/lib/runner.js (hveldstra 2022-01-06 15:05:12 +0000 213)   phaser.on('done', function () {
```

```
git log 8beb82c4
```

```
commit 8beb82c437ef61bb06207a575def5d331227dbd0
Author: hveldstra <h@veldstra.org>
Date:   Thu Aug 5 12:23:42 2021 +0100

    feat: distribute work across worker threads
```

There's also a similar bit of code for `phaseCompleted` event - don't we need to bring that back in too?

Also, is there a reason to check the `pause` property here vs reusing `isIdlePhase()`?

juanig1:
** This should be a draft **
I went through the same process, and wanted to get your input, as wasn't really sure as why stat emits were commented out

> Also, is there a reason to check the pause property here vs reusing isIdlePhase()?
`isIdlePhase` does not handle paused phases, it should, right? If you are ok I'll update it to look for a pause prop, wanted to be sure I wasn't changing 'expected' behavior

> There's also a similar bit of code for phaseCompleted event - don't we need to bring that back in too?
Wanted your input, but yeah _i think_ there is no harm and we cover cases were phase start and ends in different stats buckets

Going to test these two changes again with artillery-pro

core/lib/is-idle-phase.js
```diff
@@ -6,7 +6,8 @@ function isIdlePhase(phase) {
   return (
     (phase.arrivalRate === 0 && !phase.rampTo) ||
     phase.arrivalCount === 0 ||
-    phase.maxVusers === 0
+    phase.maxVusers === 0 ||
+    phase.pause > 0
```
juanig1, line 10: this means that now the expected amount of workers that we wait for when collecting stats is actually the ones that _are_ running.

We still need the added `emit('stats',..)` for cases were the work is divided among n workers but the last phase is an arrivalCount (which means only one active worker).

Added a test that checks we are dividing properly in scritps that mix `arrivalRate` and `arrivalCount`

PR #1662, opened 2022-11-10
Feature/lambda graviton arch as default

### LINKED ISSUE #1653: Add support for running on Graviton
[Graviton](https://aws.amazon.com/blogs/aws/aws-lambda-functions-powered-by-aws-graviton2-processor-run-your-functions-on-arm-and-get-up-to-34-better-price-performance/) is awesome:

> up to 19 percent better performance at 20 percent lower cost

With this change, Artillery's Lambdas will default to running on Graviton. Artillery itself already runs on ARM, and most custom JS should not be affected. The user will have the ability to use the legacy `x86_64` CPUs with `--platform-opt arch=x86_84` option.

juanig1:
Closes #1653

### DIFF
--- a/lib/platform/aws-lambda/index.js
+++ b/lib/platform/aws-lambda/index.js
@@ -204,7 +204,6 @@ class PlatformLambda {
     });
     artillery.log(` - Lambda function: ${this.functionName}`);
     artillery.log(` - Region: ${this.region}`);
-    artillery.log(` - Architecture: ${this.architecture}`);
     artillery.log(` - AWS account: ${this.accountId}`);
 
     debug({ bucketName, s3path, sqsQueueUrl });
--- a/lib/platform/aws-lambda/index.js
+++ b/lib/platform/aws-lambda/index.js
@@ -43,6 +43,7 @@ class PlatformLambda {
 
     const platformConfig = platformOpts.platformConfig;
 
+    this.architecture = platformConfig.architecture || 'arm64';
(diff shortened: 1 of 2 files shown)


lib/platform/aws-lambda/prices.js
```diff
@@ -1,22 +1,29 @@
 // Compute pricing for AWS Lambda
 // GB/second divided by 10e10
+// https://aws.amazon.com/lambda/pricing/
 module.exports = {
   base: {
-    x86: 1666670
```
juanig1, line 5: I think this needed to be adjusted:
Price for us-east-1 ([aws](https://aws.amazon.com/lambda/pricing/)):
`$0.0000166667` for every GB-second
`1666670 / 10 ^ 10 = 0.000166667`
If im not missing anything we were 'overcharging'

juanig1: 10e10 = 10 * 10^10

lib/platform/aws-lambda/index.js
```diff
@@ -203,6 +204,7 @@ class PlatformLambda {
     });
     artillery.log(` - Lambda function: ${this.functionName}`);
     artillery.log(` - Region: ${this.region}`);
+    artillery.log(` - Architecture: ${this.architecture}`);
```
hassy, line 207:
```suggestion
```

Don't think we need to include this here.

PR #1671, opened 2022-11-14
feat: Better error messages (command not found + cleaner logs)

juanig1:
Work a bit on a better user experience
- Add `oclif-plugin-not-found`
  - Suggest similar commands if original command input does not exist

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -36,9 +36,14 @@ async function main() {
   } catch(err) {
     await require('@oclif/errors/handle')(err);
 
-    console.log(err.stack);
-    console.log(err.code);
-    console.log(err.message);
+    let errInfo = ['stack', 'code', 'message'];
+
+    errInfo.forEach((e) => {
+      let info = err[e];
+      if (info) {
+        console.log(info);
+      }
+    });
   }
 }
 
--- a/bin/run
+++ b/bin/run
@@ -6,6 +6,7 @@
 
 const command = require('@oclif/command');
 const { createGlobalObject } = require('../lib/artillery-global');
+const debug = require('debug')('run');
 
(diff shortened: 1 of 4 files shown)


bin/run
```diff
@@ -36,9 +37,15 @@ async function main() {
   } catch(err) {
     await require('@oclif/errors/handle')(err);
 
-    console.log(err.stack);
-    console.log(err.code);
     console.log(err.message);
+
```
hassy, line 41: I suggest this is reverted. The error message printed here is important information for the user. We should not require a debug flag to be set for them to see it.

test/testcases/errors-and-warnings.bats
```diff
@@ -31,9 +31,9 @@
   [[ true ]]
 }
 
-@test "Exits with non zero when an unknown command is used" {
-    run ./bin/run makemeasandwich --with cheese
-    [[ $status -ne 0 ]]
+@test "Suggest similar commands if unknown command is used" {
+   output=echo "n" | ./bin/run helpp 2>&1 | grep -q "Did you mean help"
+   [[ $output -eq 0 ]]
```
hassy, line 36: What happens when a command that isn't similar to any existing command is used? We still need a test for that.

juanig1: it's [not configurable](https://github.com/oclif/plugin-not-found/blob/main/src/index.ts#L15) sadly. I kind of expected it to discard suggestions where dist > N, but no

hassy: Okay, cool 👍 Let's merge this!

juanig1:
moved here https://github.com/artilleryio/artillery/pull/1699

PR #1683, opened 2022-11-18
Hotfix/ramp regression

juanig1:
When a worker had a phase where `rampTo = arrivalRate = 0`, the tick was not being calculated properly, causing the run to hang.

This happened because on the latest refactor of the ramp up logic, arrivalRate and rampTo defaults values (`1` and `arrivalRate+1` respectively) were deleted, as they were causing unexpected arrivals.

The bottom line is that phases with `rampTo = arrivalRate = 0` are valid and we need to handle them properly.

The simplest test case is running a test with two rampTo phases the first one with max(rampTo, arrivalRate) < #workers, and the second one with max(rampTo, arrivalRate) < #workers, see test/scripts/ramp-regression-1682.json for an example.

This causes workers to be 'idle' (`rampTo = arrivalRate = 0`) in the first phase but active in the second one which is totally fine


hassy:
rebased on top of `main`

PR #1684, opened 2022-11-18
chore: switch to 'invoke' as 'invokeAsync' has been deprecated

### LINKED ISSUE #1681: AWS Lambda InvokeAsync API Deprecated
Received this email from AWS about a deprecation

> You are receiving this message because we identified AWS Lambda functions in your account that are using the deprecated InvokeAsync API [1], which was deprecated in 2015 and is replaced by the Invoke API [2].

> While Lambda continues to support customers using the deprecated API, the API does not receive any new updates. The deprecated InvokeAsync API does not consider reserved concurrency [3] for a function as reserved concurrency was launched in 2017, after the InvokeAsync API was deprecated. When a function using the InvokeAsync API with reserved concurrency experiences throttles, you may also experience throttles for other functions in the account. With Invoke API with InvocationType as Event, Lambda will only throttle the function with a reserved concurrency in such an event without impacting other functions in the account.

> To ensure that your functions continue to benefit from optimizations in asynchronous processing, we recommend you to migrate your existing functions to the Invoke API. If the impacted AWS Lambda functions are already deleted, no further action is required.

If you have any questions, please contact AWS Support [4].

[1] https://docs.aws.amazon.com/lambda/latest/dg/API_InvokeAsync.html
[2] https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html
[3] https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html
[4] https://aws.amazon.com/support

juanig1:
Fixes #1681 
Thank you willjk!

### DIFF
--- a/packages/artillery/lib/platform/aws-lambda/index.js
+++ b/packages/artillery/lib/platform/aws-lambda/index.js
@@ -391,11 +391,38 @@ class PlatformLambda {
     debug('Lambda event payload:');
     debug({ event });
 
-    const args = Buffer.from(JSON.stringify(event));
+    const payload = JSON.stringify(event);
 
-    await lambda.invokeAsync({
+    // Wait for the function to be invocable:
+    let waited = 0;
+    let ok = false;
+    while (waited < 120 * 1000) {
+      try {
+        var state = (await lambda.getFunctionConfiguration({ FunctionName: this.functionName}).promise()).State;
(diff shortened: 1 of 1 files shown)


juanig1:
Would wait until lambda executions are fixed to test its working after the rebase

PR #1697, opened 2022-11-24
refactor: turn this into a monorepo

hassy:
This is a big PR but most of it is shuffling existing code around.

Summary: we're turning this repo into a monorepo. 🧱 

- We are now using npm workspaces to manage different parts of the codebase
- There are two internal packages, and one external package called `artillery` that makes use of the two internal packages.
  - "Internal" packages are not meant to be used by any third-party code directly. We will work on the assumption that **only code in this repo** uses those packages. We're not going to be following semver for them.

Turbo is used as just a test runner right now. Because the test suite is now split up between two packages and more of it runs in parallel, we're down to ~10m from ~15m for the suite to run. 🏎️


hassy:
@SocketSecurity ignore turbo@1.5.1

PR #1689, opened 2022-11-21
fix: wildcard added to catch socket events

somanwal:
Support of this package https://www.npmjs.com/package/socketio-wildcard given only for socket V1 and V2 but it's also working for socket V3 and V4. Without this package we can't catch event with socket.on('*', callback).
Tested this on my machine.

PR #1700, opened 2022-11-25
Feature/socketio wildcard

juanig1:
Original PR: https://github.com/artilleryio/artillery/pull/1689
Update with main, add test for reported original issue (getting event if using wildcard in a custom processor)

Update:
Original PR and issue are stale; bringing back wildcard logic from [here](https://github.com/artilleryio/artillery/commit/729ac3e378d5d6b0dd3ec77cceb6807322cfd5f3).
- Added problematic user case as cli test 
- Reused core's simple socketio server
- Using * instead of onAny fixes original 'passive listening' test
  - seems wildcard is not compatible with onAny?
  - Are we ok with handling events this way? 
  - From what im reading end result should be the same


packages/artillery/test/runner.sh
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
 echo "# starting target"
-# &>/dev/null PORT=3003 node ./test/core/targets/simple.js &
-target_pid=$!
 &>/dev/null node ./test/gh_215_target.js &
+target_pid=$!
+&>/dev/null node ../core/test/core/targets/simple_socketio.js &
```
juanig1, line 6: reuse core's simple socketio server

hassy:
Rebased on `main`

PR #1703, opened 2022-11-28
docs: add simple development doc

juanig1:
Simple development section in the Readme


hassy:
We already have a contributor's guide here: https://github.com/artilleryio/artillery/blob/main/CONTRIBUTING.md

The stuff that's not already covered there (like commitlint, creating an issue first) should probably go there instead of the README.

We can link to `CONTRIBUTING.md` from the README.

PR #1724, opened 2022-12-07
fix(dist): distribute maxVusers value to active workers only

juanig1:
See original report [here](https://github.com/artilleryio/artillery/discussions/1711)

`maxVusers` are less than expected for cases were one or more workers are idle (for example: `arrivalCount` is being used or `arrivalRate` is less than `#workers`)
This is caused because it is being distributed across all workers not only active ones for that phase.

This pr changes `maxVusers` distribution to only take into account workers that generate arrivals.

This means the total `maxVusers` value is going to be respected, but we still can have odd distributions based on expected arrivals, max virtual users, and workers.

For example: `arrivalRate=10, maxVusers=10, workers=7` is not ideal
image


hassy:
super!

PR #1734, opened 2022-12-14
fix(lamda): replace hardcoded memorySize with actual parameter

juanig1:
Lambda:
Now follows `memory-size` obtained from user (still defaults to 4096 if no option present).

```
artillery run
 --platform aws:lambda 
 ...
 --platform-opt memory-size=2048 hello.yaml
```
Results in the following lambda properties:
image


mikevercoelen:
Awesome, @hassy any idea when this gets released?

PR #1738, opened 2022-12-15
Fix test suite

juanig1:
- Fix turbo
Plugin-expect:
  - Fix tests not running (mostly paths?)
  - Fix failing tests from using console.log instead of artillery.log

<!-- Closes ART-725 --!>


hassy:
🎉

PR #1765, opened 2023-01-12
feat(engines): add cleanup hook support for custom engines

juanig1:
Support a cleanup function (`cleanup`) on custom engines to allow gracefully shutting down.
Example: Posthog engine should shutdown the client to force flushes https://github.com/artilleryio/artillery-engine-posthog/pull/1/commits/f8de8118729f849f73dc30e887bd222c983b9e10?w=1#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R22

### DIFF
--- a/packages/core/lib/runner.js
+++ b/packages/core/lib/runner.js
@@ -215,13 +215,24 @@ function run(script, ee, options, runState, contextVars) {
   phaser.on('done', function () {
     debug('All phases launched');
 
-    const doneYet = setInterval(function checkIfDone() {
+    const doneYet = setInterval(async function checkIfDone() {
       if (runState.pendingScenarios === 0) {
         clearInterval(doneYet);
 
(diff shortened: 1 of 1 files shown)


packages/core/lib/runner.js
```diff
@@ -223,6 +224,16 @@ function run(script, ee, options, runState, contextVars) {
 
         const totals = SSMS.pack(intermediates);
 
+        runState.engines.map((engine) => {
+          if (engine.cleanup) {
+            try {
+              engine.cleanup().then(() => debug('engine unloaded:', engine.__name)) ;
```
hassy, line 230: This should be refactored to use async/await. Otherwise there's no guarantee that a given `cleanup()` handler finishes before the CLI exits, or that they complete in the order they're being called.

PR #1767, opened 2023-01-13
feat: add usage ping if no telemetry is enabled

juanig1:
We should disable ip capturing on posthog prior to merging this
```Project Settings -> Data capture configuration -> Discard client IP data```

### DIFF
--- a/packages/artillery/lib/telemetry.js
+++ b/packages/artillery/lib/telemetry.js
@@ -41,31 +41,36 @@ const init = () => {
     shutdown: noop
   };
 
-  if (telemetryDisabled) {
-    return telemetry;
-  }
-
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/artillery/lib/telemetry.js [hunk 1]
```diff
@@ -41,31 +41,36 @@ const init = () => {
...
-      eventPayload.properties = Object.assign(
-        eventPayload.properties,
-        telemetryDefaults
-      );
+      let eventPayload;
+
+      if (telemetryDisabled) {
+        eventPayload = {
+          event,
+          distinctId: 'artillery-core-no-telem'
```
hassy, line 51: Let's keep this as `artillery-core`

PR #1777, opened 2023-01-19
fix: update 'commons' dependency to '@artilleryio/int-commons'

aukevanleeuwen:
To be honest I'm not 100% sure this is the right change, but when installing artillary globally (in my CI), it errors out with 
```
error Couldn't find any versions for "commons" that matches "*"
```

I cannot imagine that you really wanted to load [this library](https://www.npmjs.com/package/commons), but rather reference your own.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -3,6 +3,7 @@
   "version": "2.0.2",
   "main": "./index.js",
   "dependencies": {
+    "@artilleryio/int-commons": "latest",
     "@artilleryio/sketches-js": "^1.0.4",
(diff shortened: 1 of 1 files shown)


hassy:
Thanks @aukevanleeuwen! yep the test failure is unrelated (we're still updating the test suite after a recent conversion to a monorepo). They should be ok on `main`

aukevanleeuwen:
Well that was fast. :-) I'll be awaiting the release.

dogfish182:
Hi there, Any ideas on when this will release?

Issue #1751, opened 2022-12-23
Check if report folder exists before starting test [improvement]


capan:
Version info: 
```
VERSION INFO:

Artillery Core: 2.0.0-27
Artillery Pro:  not installed (https://artillery.io/product)

Node.js: v16.17.1
OS:      darwin
```

Running this command: 
```
 artillery run --output result/ts-lambda-report.json test.yml
```

I expected to see this happen:

*Return file not found error before running the test.yml.*

Instead, this happened:
<img width="950" alt="image" src="https://user-images.githubusercontent.com/7265736/209366167-56d01e69-2b19-4894-b2ef-482ee6d85969.png">
*I have a rather long test case that lasts ~10 minutes and I want a report after the test run.  I think it would be better to generate or validate file location exists before running the test not after the test finished.*

Files being used:
```
<relevant yaml/js/csv go here>
```

PR #1783, opened 2023-01-25
Issue 1751 check report folder

InesNi:
## What
I've moved the validation of directory path provided with --output flag from after to before the test runs.
This was done to address the #1751 issue.

## Why
The path provided with -o flag is better checked before test runs as this saves user time when path is nonexistent, especially with long running test cases.
This way the error stops the test from running at all if not being able to output report rather than only after the test has completed.

## How
Created the checkDirExists function that:
- Checks if path is a file to make sure to only use the directory path (if file doesn't exist but path to its directory does, file with provided name is automatically created so we only check directory)
- Checks if directory path exists with fs.existsSync()
- Throws a new Error object with "No such directory!" message if path does not exist.

This function is then called before test run inside of an if block so it's triggered only when both -o and path are provided.

## Testing
I wrote a test for the new functionality that checks that:
- It exits with error when -o path provided is nonexistent (by checking the exit code)
- The correct error message is displayed (checking stdout includes the error message)
- It exits before test run ( checking stdout does not include "Summary report" as that outputs after the test run, before the report is created)


packages/artillery/lib/cmds/run.js
```diff
@@ -529,6 +536,20 @@ async function sendTelemetry(script, flags, extraProps) {
   }
 }
 
+function checkDirExists(output) {
+  if (!output) {
+    return;
+  }
+  // If destination is a file check only path to its directory
+  const exists = path.extname(output)
```
hassy, line 544: Relying on `extname()` to check whether a path points to a file or a folder is not ideal as it will break on directories with `.` characters in them. E.g. if the user passes `-o my.reports/` to Artillery, the result of `path.dirname()` will be "." rather than `my.reports`.

[`fs.lstat().isDirectory()`](https://nodejs.org/api/fs.html#fslstatpath-options-callback) would be a more reliable way to check whether a path points to a directory or a file. Once we know which one it is, we can check if the directory exists.

packages/artillery/lib/cmds/run.js
```diff
@@ -529,6 +536,20 @@ async function sendTelemetry(script, flags, extraProps) {
   }
 }
 
+function checkDirExists(output) {
+  if (!output) {
```
hassy, line 540: I don't think we need this check here because this function will be called only if `flags.output` is truthy already.

packages/artillery/lib/cmds/run.js
```diff
@@ -529,6 +536,20 @@ async function sendTelemetry(script, flags, extraProps) {
...
+    return;
+  }
+  // If destination is a file check only path to its directory
+  const exists = path.extname(output)
+    ? fs.existsSync(path.dirname(output))
+    : fs.existsSync(output);
+
+  if (!exists) {
+    throw new Error('No such directory!');
```
hassy, line 549: Would be nice to catch this error and print a user-friendly message, e.g. `Directory x/y/z does not exist`.

hassy:
This is great, thank you @InesNi! Left a couple of suggestions on the PR. The failing test is not related to the changes in this PR so we'll be good to merge.

packages/artillery/lib/cmds/run.js
InesNi: Thank you for the review and suggestions @hassy !  I see now how using path.extname() could be problematic.

The reason I didn't use `fs.lstat().isDirectory()` or `fs.statSync().isDirectory` (which is used in the original getLogFileName function) in the first place, is because these functions return the `fs.Stats` object (and therefore the `isDirectory()` method) only if the path provided actually exists. If the path does not exist  error is thrown (or returned in case of `fs.lstat()` ).
Due to this unfortunately I could not find a way to use them to find out whether a path is a file or a directory prior to checking its existence. 
But I am new at this and could be missing something..

We could maybe do something like this:

```
function checkDirExists(output) {

  // We check first if parent directory path is valid
  const parent = fs.existsSync(path.dirname(output));
  const noDirError = new Error(`Directory ${output} does not exist.`);

  if (!parent) {
    throw noDirError;
  }

  // If parent is valid we check if the whole path is valid
  // If the path does not exist and the last character of path is a path separator
  // then it is considered a nonexistent directory
  if (!fs.existsSync(output) && output.at(-1) === (path.sep || '/')) {
    throw noDirError;
  }
}
```
I believe this might screen the output string the proper way. What do you think of something similar to above?

hassy:
I'll merge this - the error message and checking for ".json" suffix can be done in a separate PR.

cunhabruno:
It would be good if Artillery created the folder structure if it does not exist.

Issue #1522, opened 2022-07-19
Expect Plugin Feature Request


eliezercazares:
Current assertions are similar to:

"expect Variable to be equal to".. i.e.

```
expect:
  - statusCode:
    - 200
    - 301
```

However, I have a scenario where the API will create an entity and return 200, but if I send the exact same request again, API will reply with custom errors, like 403 or 500 to indicate it's a duplicate request.

I am looking to do something similar to:

```
expect:
  - statusCode:
    differentThan:
      - 200
      - 301
```

hassy:
interesting use case @eliezercazares, that check should be very straightforward to implement too!

PR #1793, opened 2023-01-28
Issue 1522 expect plugin feature

InesNi:
## What
I've added an additional expectation case/keyword for expect plugin named `notStatusCode`. It checks if response status code is different than the one/s specified.
This was done to address the #1522 in the issue tracker.

## Why
In some cases users want to test that the returned status code is different than certain status code/s, rather than equal to one.

## How
It was done by creating a new `expectNotStatusCode` function in the file where other expectations are created, and exporting it with them. It follows the same 'template' as other expectation functions - returns the `result` JS object that has `expectation`, `ok`, `expected` and `got` properties. 

Since this function is very similar to `expectStatusCode` one, the code is repetitive and could be extracted into a separate function that would be used by both `expectStatusCode` and `expectNotStatusCode` (as was done with `expectHasProperty` and `expextNotHasProperty` [here](https://github.com/artilleryio/artillery/blob/28975a7b41672189821507975719e4319ca45aab/packages/artillery-plugin-expect/lib/expectations.js#L196) ). 
However because these two functions are fairly short and readable, and the extracted function would only be used by them, I've opted not to do this as I believe it might be more readable as is. I do have the other version made and ready to commit though, if that would be the preferred solution. 

## Testing
I have written a test as well but unfortunately could not check if it is passing because, at this moment, the tests are failing as is (without any changes from my side) when I run them. I have however tested the new expectation manually and I believe it  is behaving as it should. 
The test itself could use a look.

Both the test and the function are positioned right under their `statusCode` counterparts for easy comparison.

### DIFF
--- a/packages/artillery-plugin-expect/lib/expectations.js
+++ b/packages/artillery-plugin-expect/lib/expectations.js
@@ -11,6 +11,7 @@ const _ = require('lodash');
 module.exports = {
   contentType: expectContentType,
   statusCode: expectStatusCode,
+  notStatusCode: expectNotStatusCode,
   hasHeader: expectHasHeader,
   headerEquals: expectHeaderEquals,
   hasProperty: expectHasProperty,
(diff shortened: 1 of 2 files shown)


hassy:
Looks good @InesNi! thank you for the PR

The test suite is in a bit of a state right now. hoping to complete the migration from Circle CI to GitHub Actions soon.

PR #1815, opened 2023-02-10
fix: y-axis title for counter charts

chinh-tran:
Show `count` instead of `ms` as y-axis title for counter charts in HTML reports


hassy:
thank you @chinh-tran!

Issue #1741, opened 2022-12-16
Template in query string not working ?


abcfy2:
Version info: 
```
❯ npx artillery version

        ___         __  _ ____
  _____/   |  _____/ /_(_) / /__  _______  __ ___
 /____/ /| | / ___/ __/ / / / _ \/ ___/ / / /____/
/____/ ___ |/ /  / /_/ / / /  __/ /  / /_/ /____/
    /_/  |_/_/   \__/_/_/_/\___/_/   \__  /
                                    /____/


VERSION INFO:

Artillery Core: 2.0.0-27
Artillery Pro:  not installed (https://artillery.io/product)

Node.js: v18.12.1
OS:      linux
```

Running this command: 
```
npx artillery run -o result.json artillery-config.yml
```

I expected to see this happen:

Here is the example:

```yaml
  - name: "GET /mint-request/:address"
    flow:
      - get:
          url: "/mint-request/{{ derivativeAddress }}"
          gzip: true
          headers:
            Authorization: "Bearer {{ JWT }}"
            "Content-Type": "application/json"
          qs:
            network: "{{ network }}"
            email: "{{ email }}"
            requestStatus: "{{ requestStatus }}"
```

Instead, this happened:

```
❯ DEBUG=http npx artillery run -s -o result.json artillery-config.yml
Phase started: unnamed (index: 0, duration: 1s) 10:40:18(+0800)

Phase completed: unnamed (index: 0, duration: 1s) 10:40:19(+0800)

⠸ 2022-12-16T02:40:21.350Z http request: {
  "url": "https://*.*.*.*/mint-request/*****",
  "method": "GET",
  "headers": {
    "user-agent": "Artillery (https://artillery.io)",
    "authorization": "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyIjoiZHVtbXkiLCJpYXQiOjE2NDM1MTYyMjZ9.QeDmD2FNLQbrp52Bn7eCiztoQkXuBNIzWricMscop1M",
    "content-type": "application/json"
  },
  "qs": "network=%7B%7B%20network%20%7D%7D&email=%7B%7B%20email%20%7D%7D&requestStatus=%7B%7B%20requestStatus%20%7D%7D"
}
```

Seems template in `qs` is not working ?

abcfy2:
Seems this BUG has been fixed by https://github.com/artilleryio/artillery/pull/699

But why I still got this issue ?

[...]

PR #1817, opened 2023-02-14
fix: show templated qs in debug output

InesNi:
## What
This PR fixes issue #1741. 

The `requestInfo` object which is logged by the `debug` function was not templating the querystring. 

## How
I have simply added the templating function to the part of the code that checks for the existence of `qs` in order to add it to `requestInfo`. This templating function is exactly the same that gets used for templating the main code.

## Testing
Because this was just an output from debug, I have not added an automated test case for it. 

I have recreated the setup from the issue above locally and verified that `DEBUG=http` now outputs the templated string correctly.


hassy:
thank you @InesNi!

PR #1826, opened 2023-02-16
fix: Cookies will work when url is a $loopElement

### LINKED ISSUE #1824: Cookies don't work if the url is a $loopElement
Version info: 
```

        ___         __  _ ____
  _____/   |  _____/ /_(_) / /__  _______  __ ___
 /____/ /| | / ___/ __/ / / / _ \/ ___/ / / /____/
/____/ ___ |/ /  / /_/ / / /  __/ /  / /_/ /____/
    /_/  |_/_/   \__/_/_/_/\___/_/   \__  /
                                    /____/


VERSION INFO:

Artillery: 2.0.0-29
Node.js:   v18.13.0
OS:        darwin
```

Running this command: 
```
artillery run test-config.yml
```

I expected to see this happen:

When you set a cookie like shown in the `test-config.yml` we should see it being set on the actual request.  You should be able to see it [here](https://requestbin.myworkato.com/wsvvhgws).

Instead, this happened:

No cookie header in the request bin response.

Files being used:
```
# test-config.yml
config:
  target: "https://requestbin.myworkato.com"
scenarios:
  - flow:
    - loop:
      - get:
          url: "{{ $loopElement }}"
          cookie:
            saved: "tapir,sloth"
      over:
        - "/wsvvhgws"
          
```

I have a fix for this, but don't have a test to simulate this yet.  The reason is that the cookie is apparently set on just the last portion of the url(i.e without the domain).

anoopknayak:
The issue was that we were computing the url after setting the cookies if it was templated. This led to the cookies being set on '' domain in the cookie jar leading to it not being set on the request properly. The fix is to move the logic of url computation before setting the cookies. Tests are also added for the same so that we can detect if it would break in future due to code changes.

Fixes #1824

### DIFF
--- a/packages/core/lib/engine_http.js
+++ b/packages/core/lib/engine_http.js
@@ -399,6 +399,20 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
         );
         requestParams.headers = templatedHeaders;
 
+        // We compute the url here so that the cookies are set properly afterwards
+        let url = maybePrependBase(
+          template(requestParams.uri || requestParams.url, context),
+          config
+        );
+
+        if (requestParams.uri) {
+          // If a hook function sets requestParams.uri to something, request.js
+          // will pick that over .url, so we need to delete it.
+          delete requestParams.uri;
+        }
+
+        requestParams.url = url;
+
         if (
           typeof requestParams.cookie === 'object' ||
           typeof context._defaultCookie === 'object'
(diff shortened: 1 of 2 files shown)


hassy:
thank you @anoopknayak! tests passing on Node.js v16 - the failures on other versions are not related to code changes (looks like we need to give GHA runners more memory).

Issue #1221, opened 2021-11-11
CSV data is duplicated across threads


hassy:
Every worker thread gets its own copy of the entire payload - that means that the default `sequence` traversal through the rows is broken as every row will be reused by every worker.

hassy:
We won't be fixing this, as it introduces the requirement of inter-VU coordination / global state, which is especially tricky to solve in distributed tests. The recommendation is to use `random` for load tests, and only use `sequence` for tests run in "solo" mode with just one VU.

MagicKriss:
Hi, @hassy !
I stumbled upon this issue while running stress tests.
As you mentioned, a "proper" implementation would require some global state. 

However, while trying to find a possible workaround, I dug through the source code and noticed that the whole payload data is being copied to each worker. 
In this process, the same as phases, payload data could be distributed among workers similarly. 

What do you think about this approach? I could create a PR with the implementation.

PR #1829, opened 2023-02-18
fix: Distribute csv payload across workers

MagicKriss:
Ref #1221

This is a fix for a bug where the payload would get copied for each worker, and therefore each CSV row will get duplicate calls.
In my case, it meant that, when running artillery on unique historical data, it would create 1 new request and hit cache from the rest of the workers.

It would be nice to have this bug at least documented in docs, but I noticed that arrival rates are already divided between workers, so in the same place, the payload could be divided as well.

I have split this task into 2 commits:
* First refactoring the `diviedWork` functionality so that it's more readable and adding the ability to more easily update it without changing it's behavior.
* Then, adding the functionality for dividing payload between workers.

### DIFF
--- a/packages/artillery/lib/dist.js
+++ b/packages/artillery/lib/dist.js
@@ -11,127 +11,135 @@ const isIdlePhase = require('@artilleryio/int-core').isIdlePhase;
 module.exports = divideWork;
 
 /**
+ *
  * Create a number of scripts for workers from the script given to use by user.
+ *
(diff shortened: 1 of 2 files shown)


hassy:
thanks for the PR @MagicKriss!

hassy:
aaaand merged, sorry for the delay. Thank you for contributing to Artillery! @MagicKriss

PR #1868, opened 2023-03-14
feat: new relic publish metrics plugin target

InesNi:
## What
New Relic target for `artillery-publish-metrics` plugin. Supports sending metrics to New Relic through **Metric API**. 

## How
`NewRelicReporter` class that has the plugin `config` from test script and `events` as properties. On every `'stats'` event it takes metrics from the `stats` object, formats and packs them into `HTTP POST` request in format recognised by New Relic Metric API, and sends the request using `got`  library. 

Configuration options for users are:

- `licenseKey` -- license key must be provided,
- `region` -- two regions available for New Relic, `us` and `eu`, defaults to `us`. Each region has its own endpoint so it is required to set region as `eu` if users New Relic account hosts data in the EU data center.
- `prefix` -- prefix for metric names created by Artillery; defaults to `artillery.`
- `attributes` -- a list of name:value strings to use as attributes/tags for all metrics sent during a test. 
- `excluded` -- A list of metric names which should not be sent to New Relic. Defaults to an empty list, i.e. all metrics are sent to New Relic.
- `includeOnly` -- A list of specific metrics to send to New Relic. No other metrics will be sent. Defaults to an empty list, i.e. all metrics are sent to New Relic.

The class is then initialised with the `new` constructor inside `createNewRelicReporter` function which is exported, and plugged into the main artillery-plugin-publish-metrics `index.js` with other target reporters. 

The `README.md` has been updated by adding New Relic to targets list and removing it from wish list. 
There is also a New Relic specific `README.md` file inside the newrelic folder that has a plugin description consistent with the descriptions used for other targets in [artillery.io](https://www.artillery.io/docs/guides/plugins/plugin-publish-metrics#overview)

## Testing
As no testing has been set yet (from what I could see) for the targets in this plugin, I did not write specific tests yet. I did see that `config.yaml` files were added for each target so to reflect that I created an equivalent file for the New Relic target.

I have however tested the plugin manually with different scripts and combinations to make sure everything is created/ formatted properly and that  the metrics arrive as they should to New Relic and in those tests I did not notice any issues. 

## Screenshots

Screenshot from 2023-03-14 16-48-54
Screenshot from 2023-03-14 16-50-06

### DIFF
--- a/packages/artillery-plugin-publish-metrics/README.md
+++ b/packages/artillery-plugin-publish-metrics/README.md
@@ -11,9 +11,10 @@ The plugin sends metrics and events from Artillery tests to external monitoring 
 - [Prometheus](https://prometheus.io/docs/concepts/metric_types/) via [Pushgateway](https://prometheus.io/docs/instrumenting/pushing/)
 - [Honeycomb](https://honeycomb.io)
 - [Lightstep](https://lightstep.com)
+- [New Relic](https://newrelic.com/)
 - [Mixpanel](https://mixpanel.com)
 - InfluxDB with [Telegraf + StatsD plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/statsd)
-- StatsD
+- StatsD 
 
 ## Docs
 
@@ -21,7 +22,6 @@ The plugin sends metrics and events from Artillery tests to external monitoring 
 
 ## Wishlist
 
-- NewRelic
 - InfluxDB (HTTP API)
 - Splunk
 - ELK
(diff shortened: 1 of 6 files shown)


hassy:
Looks great @InesNi, thank you so much for the contribution!

InesNi:
Just as a side note, this was not tested with Artillery PRO (as I do not have it), only with `artillery run`. I believe it should still work, but should probably be tested for good measure.

hassy:
Just tested it, works like a charm!

Thank you for including a comprehensive `README`, I'll move that into our docs (that repo is not open source yet).

InesNi:
That's great! Thank you @hassy !

 I was actually just about to submit the PR for the Splunk target as well, but noticed their SDK now has several high and critical vulnerabilities amongst other issues I encountered along the way. 
Due to this I have decided to drop the SDK and instead use the HTTP API directly.

PR #1872, opened 2023-03-16
feat: add support for Artillery Cloud

hassy:
This PR adds support for sending test results to the Artillery dashboard.

(Artillery Cloud is in a closed beta right now. Sign up for the waiting list here: https://www.artillery.io/cloud)


hassy:
Note: something strange is going on with the `rampTo` test failure. There are no changes in this PR that could've affected how phases work, and while the test is passing on `main` when run by GitHub Actions, it fails for me in the same way when running both locally and in a fresh Codespace. cc @juanig1

PR #1877, opened 2023-03-20
feat: splunk publish metrics plugin target

InesNi:
## What
Splunk target for artillery-publish-metrics plugin. Supports sending metrics to Splunk Observability Cloud via Ingest API.

## How
SplunkReporter class that has the plugin config from test script and events as properties. On every 'stats' event it takes metrics from the stats object, formats and packs them into a format recognised by the ingest API, and sends the request using  `got` library. 

### Configuration options for users are:

`accessToken` -- org ingest access token must be provided,
`realm` --  defaults to `us0`. Must be set if the account's realm is different than `us0` as each realm has its own endpoint.
`prefix` -- prefix for metric names created by Artillery; defaults to artillery.
`dimensions` -- a list of name:value strings to use as dimensions for all metrics sent during a test.
`excluded` -- A list of metric names which should not be sent to Splunk. Defaults to an empty list, i.e. all metrics are sent.
`includeOnly` -- A list of specific metrics to send to Splunk. No other metrics will be sent. Defaults to an empty list, i.e. all metrics are sent.
The class is then initialised with the `new` constructor inside `createSplunkReporter` function which is exported, and plugged into the main artillery-plugin-publish-metrics `index.js` with other target reporters.

The `README.md` has been updated by adding Splunk to targets list and removing it from wish list.
There is also a Splunk specific `README.md` file inside the splunk folder that has a plugin description consistent with the descriptions used for other targets in [artillery.io](https://www.artillery.io/docs/guides/plugins/plugin-publish-metrics#overview)

## Note:
At first I used the Splunk (SignalFx) Node.js SDK as per convention, but I ran into multiple issues with it:

- Installing versions 8.x.x-beta.x of the SDK would run into a webpack error `npm ERR! sh: 1: webpack: not found`. This error was inconsistent. In some versions (e.g. 8.0.0-beta.4) it was present, in others it was not;
- In some of the versions where the error above was not present, high and critical security vulnerabilities were present instead;
- Furthermore, these versions appeared released on NPM, but there was no release on Github or mention of the change.
- I ended up installing version 7.4.1. which seemed stable at the time, but yesterday presented with multiple high and one critical vulnerability
- The SDK did not have a lot of documentation;
- The error handling of the SDK was not the best. For example, even the success case just returns OK, and when it errors there is not much information coming from the error.
- Also for the functionality I am using, the SDK anyway doesn't add much.

Given all of the above, I decided to call the HTTP API directly instead.

## Testing
For now I have created a config.yaml file for the Splunk target equivalent to the ones set for other targets, and have tested the plugin manually. Everything seems to be working as it should, metrics get formatted correctly and successfully ingested by Splunk API.

## Screenshots
Screenshot from 2023-03-20 16-37-32

image

### DIFF
--- a/packages/artillery-plugin-publish-metrics/README.md
+++ b/packages/artillery-plugin-publish-metrics/README.md
@@ -8,6 +8,7 @@ The plugin sends metrics and events from Artillery tests to external monitoring 
 
 - [AWS CloudWatch](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/working_with_metrics.html)
 - Datadog - (via [agent](https://docs.datadoghq.com/agent/) or [HTTP API](https://docs.datadoghq.com/api/))
+- [Splunk](https://splunk.com)
 - [Prometheus](https://prometheus.io/docs/concepts/metric_types/) via [Pushgateway](https://prometheus.io/docs/instrumenting/pushing/)
 - [Honeycomb](https://honeycomb.io)
 - [Lightstep](https://lightstep.com)
@@ -23,5 +24,4 @@ The plugin sends metrics and events from Artillery tests to external monitoring 
 
 - NewRelic
 - InfluxDB (HTTP API)
-- Splunk
 - ELK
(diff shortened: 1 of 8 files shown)


hassy:
Great work @InesNi! Love how comprehensive your PR descriptions are!

Apologies for taking so long to merge. Going out in the next release of the CLI! 🎉

PR #1880, opened 2023-03-21
test(ramp): add test for 1s duration ramp

juanig1:
test for https://github.com/artilleryio/artillery/pull/1878

### DIFF
--- a/packages/artillery/test/cli/command-run.test.js
+++ b/packages/artillery/test/cli/command-run.test.js
@@ -26,10 +26,7 @@ tap.test(
       '-o',
       'totally/bogus/path'
     ]);
-    t.ok(
-      exitCode !== 0 &&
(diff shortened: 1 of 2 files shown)


packages/artillery/test/cli/command-run.test.js
```diff
@@ -288,6 +285,47 @@ tap.test(
...
+
+    const reportSingleFile = 'single_worker.json';
+    const reportSingleFilePath = await getRootPath(reportSingleFile);
+
+    const [exitCodeMultiple] = await execute([
+      'run',
+      '-o',
+      reportMultipleFile,
+      'test/scripts/ramp-regression-1682.json'
+    ]);
```
hassy, line 306: Should this set `WORKERS` explicitly to avoid false positives on machines with only one CPU?

PR #1886, opened 2023-04-04
feat: add initial version of apdex plugin

hassy:
Add support for Apdex [1] scores to Artillery.

- Threshold may be set via `config.apdex.threshold`
- Each HTTP request is compared against the threshold and classified as "satisfied", "tolerated" or "frustrated"
- Total number of each type of request is reported via `apdex_satisfied`, `apdex_tolerated`, and `apdex_frustrated` metrics
- Apdex score is calculated and reported at the end of the test run, along with a summary ("good" / "fair" / "poor")
- `ensure` checks may use the `apdex` variable in threshold & condition expressions

1. https://en.wikipedia.org/wiki/Apdex

### DIFF
--- a/packages/artillery-plugin-apdex/index.js
+++ b/packages/artillery-plugin-apdex/index.js
@@ -39,6 +39,11 @@ class ApdexPlugin {
     global.artillery.ext({
       ext: 'beforeExit',
       method: async (testInfo) => {
+        if (typeof this.script?.config?.apdex === 'undefined' ||
+          typeof process.env.ARTILLERY_DISABLE_ENSURE !== 'undefined') {
+          return;
+        }
+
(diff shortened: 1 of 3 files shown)


packages/artillery-plugin-apdex/index.js
```diff
@@ -0,0 +1,82 @@
...
+
+          global.artillery.apdexPlugin = {
+            apdex: apdexScore,
+            ranking
+          };
+
+          console.log(`\nApdex score: ${apdexScore} (${ranking})`);
+        } else {
+          // no data
+        }
```
manfromanotherland, line 74: Does this need to be here?

manfromanotherland (commented):
🚀

PR #1947, opened 2023-06-14
Feat: Add new configuration to metrics-by-endpoint

### LINKED ISSUE #1945: Metrics-By-Endpoint plugin not adhereing to useOnlyRequestNames
When using the metrics-by-endpoint plugin with `useOnlyRequestNames` set to `true`, if no request name is set the requestName sent in the metric is then sent as the full request path.

**Version info:**
```
Artillery: 2.0.0-33
Node.js:   v17.8.0
OS:        darwin
```

**I expected to see this happen:**

Based on the name of the plugin config setting(`useOnlyRequestNames`), if this is set to `true` I would expect only request names to be used and if one is not provided the metric to be skipped.

If this config setting is set to `false`, then I would expect the existing logic to continue in which the requestName is `{requestPath} + {requestName}` or simply just `{requestPath}`

**Instead, this happened:**
When the `useOnlyRequestNames` setting is set to true, if no name is specified for the request, the request path is instead sent.

cjsaurusrex:
Resolves #1945

Adds 2 new configuration options for the `metrics-by-endpoint` plugin; `stripQueryString` and `ignoreUnnamedRequests`. Both of these should help in reducing the number of unique metrics being produced by the plugin for larger more complex test suites.

### DIFF
--- a/packages/artillery-plugin-metrics-by-endpoint/README.md
+++ b/packages/artillery-plugin-metrics-by-endpoint/README.md
@@ -70,6 +70,14 @@ scenarios:
 
 This is due to different naming patterns for metrics in both plugins.
 
+## Configuration
+| Name | Default | Description |
+|------|---------|-------------|
+| useOnlyRequestNames | `false` | If `true`, will use only the `name` field of a request as the metric name. If `false`, the `name` will be prepended to the standard metric name if it exists. |
+| stripQueryString | `false` | If `true`, query-strings will be stripped from the metric name. |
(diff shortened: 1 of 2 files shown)


packages/artillery-plugin-metrics-by-endpoint/README.md
```diff
@@ -70,6 +70,14 @@ scenarios:
 
 This is due to different naming patterns for metrics in both plugins.
 
+## Configuration
```
hassy, line 73: thanks for updating the docs!

hassy (approved):

hassy:
thank you @cjsaurusrex!

PR #1957, opened 2023-06-30
fix(skytrace): improve handling of JSON responses

hassy:
- Parse JSON responses and pass them as objects to expectation functions
- Ignore JSON parsing errors unless we actually need to use the response (e.g. to pretty-print or for a JSON-specific expectation)

### DIFF
--- a/packages/skytrace/src/commands/ping.ts
+++ b/packages/skytrace/src/commands/ping.ts
@@ -433,31 +433,44 @@ class PingCommand extends Command {
       const isJSON = contentType.match(/json/gi);
       const isXML = contentType.match(/html/gi) || contentType.match(/xml/gi);
 
-      if (flags.showBody) {
-        let language;
-        if (isJSON) {
-          language = 'json';
+      let body = context.vars.body;
+      let isBodyValidJSON = false;
+      if (isJSON) {
+        // Try to parse the body as JSON. Errors are ignored because as a
+        // general purpose HTTP client we don't want to fail on invalid
+        // JSON by default - only if the user explicitly or implicitly
+        // specifies that the body needs to be valid.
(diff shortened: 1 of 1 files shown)


packages/skytrace/src/commands/ping.ts
```diff
@@ -433,31 +433,44 @@ class PingCommand extends Command {
...
+              this.log(highlight(JSON.stringify(body, null, 4), { language: 'json' }));
+            }
+
+            if (isXML) {
+              this.log(highlight(context.vars.body, { language: 'html' }));
+            }
+
+            if (!isJSON && !isXML) {
+              this.log(context.vars.body);
+            }
```
bernardobridge, line 465: shouldn't this code path be outside the `if(isJson){}` statement? I don't think this ever gets run otherwise?

bernardobridge (changes requested):

packages/skytrace/src/commands/ping.ts
```diff
@@ -433,31 +433,44 @@ class PingCommand extends Command {
...
-        }
-        if (isXML) {
-          language = 'html';
+      let body = context.vars.body;
+      let isBodyValidJSON = false;
+      if (isJSON) {
+        try {
+          body = JSON.parse(context.vars.body);
+          isBodyValidJSON = true;
+        } catch (_parseErr) {
```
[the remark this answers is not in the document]
hassy: 👍 done

packages/skytrace/src/commands/ping.ts
hassy: yep good catch!

bernardobridge (approved):

PR #1958, opened 2023-06-30
feat: enable sending events to new relic

InesNi:
## What

Events feature for the New Relic reporter of `artillery-publish-metrics` plugin. Supports sending events marking the start and the end of a test to New Relic through **Event API**.

## How

To handle sending events to New Relic's Event API, the `sendEvent` method is implemented, which formats the `HTTP POST` request and utilizes the `got` library to send it. The `sendEvent` method is called when the first `phaseStarted` event occurs to send the event for the test's start. Additionally, it is also utilized within the existing `cleanup` method to send the event marking the test's end.

The `startedEventSent` property acts as a toggle, storing the state of the test. It is checked before sending the event to ensure that the start event is only sent upon the first occurrence of the `phaseStarted` event. This mechanism guarantees that the event for the test's start is sent solely at the beginning of the test, rather than at the start of each phase. Moreover, it ensures that the code related to the event feature in the `cleanup` method is executed only when the feature is actually used.

Configuration for users:

To send events to New Relic `event` needs to be set in the New Relic reporter configuration in the test script. Event configuration options are:

- `accountId` -- users New Relic account ID is required,
- `eventType` -- event's name, defaults to `Artillery_io_Test`
- `send` -- set to false to turn off the event. By default, if an event is configured, it will be sent. This option makes it possible to turn event creation on/off on the fly (e.g. via an environment variable)
- `attributes` -- optional list of `name:value` strings to use as attributes/tags for events sent during a test. By default, the following attributes are set for every event: `target: <target set in script config>`, `timestamp: <timestamp of start/end of test>` and `phase: 'Test Started' / 'Test Finished'`. Any `attributes` set by user will be sent in addition to the default ones.

```yaml
config:
  target: 'http://mytestwebsite.com'
  plugins:
    publish-metrics:
      - type: newrelic
        licenseKey: '{{ $processEnvironment.NEW_RELIC_LICENSE_KEY }}'
        event:
          accountId: '{{ $processEnvironment.NEW_RELIC_ACCOUNT_ID }}'
          eventType: 'Plugin_integration_test'
          attributes:
            - 'priority:low'
            - 'testId:myTest123'
```

The New Relic specific `README.md` file (sent in the original PR for the New Relic Reporter) has been updated to reflect the addition of the event feature and I have included it in this PR. Its style formatting is still consistent with the one used in Artillery's docs for other reporters.

## Testing

`config-newrelic-api.yaml` file for New Relic has been updated to include the event feature.

I have tested manually with different scripts and combinations to make sure everything is formatted properly and that the events arrive as they should to New Relic and in those tests I did not notice any issues.

## Screenshots
Image 28-06-2023 at 18 34

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/newrelic/README.md
+++ b/packages/artillery-plugin-publish-metrics/lib/newrelic/README.md
@@ -1,0 +1,59 @@
+## New Relic
+
+To send metrics and/or events to New Relic set `type` to `newrelic`:
+
+```yaml
+config:
+  plugins:
+    publish-metrics:
+      - type: newrelic
+        apiKey: '{{ $processEnvironment.NEW_RELIC_LICENSE_KEY }}'
+        prefix: 'artillery.'
+        attributes:
+          - 'type:soak-test'
+          - 'service:my-service'
+```
+
+By default, all Artillery metrics will be sent to New Relic. Each Artillery metric will create a custom New Relic metric, which may have an associated cost.
+
+### Configuration options
+
+- To send metrics to New Relic, set `type` to `newrelic`.
+- Set `licenseKey` to the license key for the account you want to send the metrics to
+- `region` -- `us` (default) or `eu`. Thes sets default New Relic endpoint. If your account hosts data in the EU data center set the region to eu.
+- `prefix` -- set a prefix for metric names created by Artillery; defaults to `artillery`.
+- `attributes` -- a list of `name:value` strings to use as tags for all metrics sent during a test
+- `excluded` -- a list of metric names which should not be sent to New Relic. Defaults to an empty list, i.e. all metrics are sent to New Relic.
+- `includeOnly` -- a list of specific metrics to send to New Relic. No other metrics will be sent. Defaults to an empty list, i.e. all metrics are sent to New Relic.
+- `event` -- set to send a New Relic event when the test starts/finishes.
+  - `accountId` -- your New Relic [account ID](https://docs.newrelic.com/docs/accounts/accounts-billing/account-structure/account-id/).
+  - `eventType` -- set to customize the event's name, defaults to `Artillery_io_Test`. Must be a string that is a combination of alphanumeric characters, underscores, and colons.
+  - `send` -- set to `false` to turn off the event. By default, if an event is configured, it will be sent. This option makes it possible to turn event creation on/off on the fly (e.g. via an environment variable)
+  - `attributes` -- optional list of `name:value` strings to use as attributes/tags for events sent during a test. By default Artillery sends the `target: <target set in the script config>`, `timestamp: <timestamp of start/end of test>` and `phase: 'Test Started' / 'Test Finished'` attributes. Any `attributes` set will be sent in addition to the default ones. Check character [restrictions] for attributes [here](https://docs.newrelic.com/docs/data-apis/ingest-apis/event-api/introduction-event-api/#instrument)
+
+```yaml
+config:
+  plugins:
+    publish-metrics:
+      - type: newrelic
+        apiKey: '{{ $processEnvironment.NEW_RELIC_LICENSE_KEY }}'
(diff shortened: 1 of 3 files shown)


packages/artillery-plugin-publish-metrics/test/config-newrelic-api.yaml
```diff
@@ -7,9 +7,14 @@ config:
     publish-metrics:
       - type: newrelic
         region: 'us'
-        licenseKey: "{{ $processEnvironment.NR_LICENSE_KEY }}"
+        licenseKey: '{{ $processEnvironment.NR_LICENSE_KEY }}'
```
hassy, line 10: Any particular reason to change double quotes to single quotes here?

packages/artillery-plugin-publish-metrics/lib/newrelic/index.js
```diff
@@ -113,22 +146,25 @@ class NewRelicReporter {
...
-  // Assembles metrics and info into req body format needed by NR metric API
-  createRequestBody(timestamp, interval, attributeList, metrics) {
+  parseAttributes(attributeList) {
     const parsedAttributes = {};
     if (attributeList.length > 0) {
       for (const item of attributeList) {
-        const attribute = item.split(':');
-        parsedAttributes[attribute[0]] = attribute[1];
+        const [name, ...value] = item.split(':');
+        parsedAttributes[name] = value.join(':');
```
hassy, line 154: Does New Relic allow `:` characters in attribute values?

packages/artillery-plugin-publish-metrics/lib/newrelic/README.md
```diff
@@ -0,0 +1,59 @@
...
+        apiKey: '{{ $processEnvironment.NEW_RELIC_LICENSE_KEY }}'
+        prefix: 'artillery.'
+        attributes:
+          - 'type:soak-test'
+          - 'service:my-service'
+        event:
+          accountId: '{{ $processEnvironment.NEW_RELIC_ACCOUNT_ID }}'
+          eventType: 'Artillery_load_test'
+          attributes:
+            'alertType:info'
```
hassy, line 48: This looks like it's missing `-`s to make it a list

hassy:
Looks good @InesNi! Btw the docs for the plugin live here: https://www.artillery.io/docs/reference/extensions/publish-metrics There's no need to duplicate them in the repo. (The docs repo isn't open source but we can copy the docs for new settings over. Appreciate the thorough job on documenting everything!)

packages/artillery-plugin-publish-metrics/test/config-newrelic-api.yaml
InesNi: Unfortunately It seems I had forgotten about this file when submitting the formatting changes to New Relic reporter prior to this PR. My local formatting has been set to match the prettier configuration found in Artillery's code, and it had changed the quotes automatically to single quotes. Since I only noticed after I made the PR I left it as is considering that it had inconsistent formatting before with both single and double quotes. 
I should of left a comment explaining this though.

packages/artillery-plugin-publish-metrics/lib/newrelic/README.md
InesNi: Ahh yes, can't believe I missed that, thank you.

packages/artillery-plugin-publish-metrics/lib/newrelic/index.js
InesNi: Yes, attributes can be formed with colons `:`. I have tested it and you can see an example on the left side of the screenshot I submitted, under dimensions (attributes show as dimensions in NR UI): one of the attributes submitted is the `target: "http://localhost:3003"`

InesNi:
Thank you for the review @hassy! 🤓 I will keep that in mind for the docs! Should I remove the `README.ms`'s from this and the Splunk PR's then?

Issue #1961, opened 2023-07-04
Incorrect payload path when using environments


cjsaurusrex:
When trying to load a(or multiple) payloads from a list within an `environment` using a relative path, the base path that is used is the directory of the config file loaded in using the `-c` parameter. This results in the file not being found and an error being thrown.

- If an Environment only contains a single payload, specified in a dictionary format, then the path is generated correctly.
E.g:
```yaml
...
  environments:
    staging:
      payload:
        path: "./data/my-data.csv"
        fields:
          - "id"
          - "name"
        skipHeader: true
        order: sequence
...
```
- If a list of payloads if specified outside of an `environment` the path is generated correctly.

**Version info:**
```
Artillery: 2.0.0-34
Node.js:   v20.3.0
OS:        darwin
```

**Running this command:** 
```
artillery run -e staging -c ./config/artillery-config.yaml my-test.yaml
```

**I expected to see this happen:**

I'd expect the `my-data.csv` file defined in my `staging` environments `payload` to be loaded from the `./data/my-data.csv` location.

**Instead, this happened:**

The `my-data.csv` is attempted to be loaded from `./config/data/my-data.csv` and an error is thrown, causing the test execution to be halted.

Error Details:
```
    Error: ENOENT: no such file or directory, open '/Users/christopher.orchard/Development/Testing/artillery-test/config/data/my-data.csv'
    Code: ENOENT
Error: ENOENT: no such file or directory, open '/Users/christopher.orchard/Development/Testing/artillery-test/config/data/my-data.csv'
    at Object.openSync (node:fs:592:3)
    at Object.readFileSync (node:fs:460:35)
    at readPayload (/opt/homebrew/lib/node_modules/artillery/lib/cmds/run.js:486:21)
    at prepareTestExecutionPlan (/opt/homebrew/lib/node_modules/artillery/lib/cmds/run.js:465:25)
    at async RunCommand.runCommandImplementation (/opt/homebrew/lib/node_modules/artillery/lib/cmds/run.js:196:20)
    at async RunCommand.run (/opt/homebrew/lib/node_modules/artillery/lib/cmds/run.js:49:5)
    at async RunCommand._run (/opt/homebrew/lib/node_modules/artillery/node_modules/@oclif/command/lib/command.js:43:20)
    at async Config.runCommand (/opt/homebrew/lib/node_modules/artillery/node_modules/@oclif/config/lib/config.js:173:24)
[...]

PR #1964, opened 2023-07-04
fix(artillery): update payload path when using environments

bernardobridge:
## What

Fixes https://github.com/artilleryio/artillery/issues/1961

## How

This was happening because checkConfig gets called twice when you pass in a scenario + config. In that situation, if you have `environments`, the first time `checkConfig` gets called, it correctly creates the `payload` in the root `config` and replaces those paths. 

However, it doesn't do the same to `config.environments.payload`. So the second time `checkConfig` is called, the `_.merge` overrides the payload path with the `config.environment.payload` unresolved path, causing the issue.

The PR addresses that by ensuring that it no longer gets called twice.

## Testing

This was tested with the test scripts provided in the issue (very helpful), against Artillery Local.

Two things are missing:

- [x] Verify this against non-local runs. This may still need to be accounted for, as I'm not sure if it uses the same logic, looking at `create-bom.js`;
- [x] I am still in the process of figuring out a potential automated test for this.

### DIFF
--- a/packages/artillery/test/scripts/scenario-payload-with-envs/scenario.yml
+++ b/packages/artillery/test/scripts/scenario-payload-with-envs/scenario.yml
@@ -9,7 +9,7 @@ config:
           skipHeader: true
           path: "./data/my-data.csv"
           order: sequence
-  target: https://http.cat
+  target: http://asciiart.artillery.io:8080
   phases:
     - duration: 1
       arrivalRate: 1
@@ -18,4 +18,4 @@ config:
 scenarios:
   - name: Test Scenario
     flow:
-      - log: "Successfully ran with id: {{ id }}"
+      - log: "Successfully ran with id {{ id }}"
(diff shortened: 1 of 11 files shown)


packages/artillery/test/scripts/scenario-payload-with-envs/scenario.yml
```diff
@@ -0,0 +1,21 @@
...
+  environments:
+    staging:
+      payload:
+        - fields:
+            - "id"
+            - "name"
+          skipHeader: true
+          path: "./data/my-data.csv"
+          order: sequence
+  target: https://http.cat
```
hassy, line 12: Since the test does not depend on any particular response from the server, can we change this to use one of the mock servers we already have, or our own service, e.g. http://asciiart.artillery.io:8080. I'd rather avoid a dependency on a 3rd party service in our test suite.

bernardobridge: If you look at the scenario, it doesn't actually make the HTTP call, the flow just does a log instead as that's enough to prove the scenario loads the payload. 

So the target could be anything really. I'll change it to ours though! 👍

PR #1968, opened 2023-07-05
chore: set up git hooks and linting

kettanaito:
## Why this change?

The current Git hooks doesn't work at all since it's not initialized to new repository developers. Moreover, most of the code quality automations are scoped to the `packages/artillery` while the entire monorepo should utilize consistent commit and code formatting.  

## Changes

- Git hooks are now automatically initialized for all devs getting started with the monorepo (run in the `prepare` command). 
- Commit linting, code linting, and code formatting are lifted to the root-level of the monorepo. All of those must be consistent across all the packages. Package-based overrides are also allowed if you create a respective config closer to the package itself (although, I wouldn't recommend that for commit and formatting aspects).

### Moving Prettier config to root-level `package.json`

Code formatting should be consistent across different packages in the same monorepo. I'm moving the `prettier` field to the root-level `package.json` to ensure that consistency.

> As an example of why this is important, if I go to random packages in the repo and save unchanged files, a bunch of them will auto-format based on my editor's Prettier config instead of respecting the monorepo's config (that one was nested in `packages/artillery` and Prettier won't pick that up for other packages). 

### Moving commitlint to the root-level

Git commit format should be consistent across the entire monorepo. We seem to be following conventional commit convention pretty heavily across packages here so there's no reason not to unify the commit message linting by putting the `commit-msg` Git hook on the root. 

As a part of this change, I've also noticed that we aren't linting the _current_ commit. It's totally possible to create a commit message against conventional commit rules and out commitlint as of now will not prevent it in any way. 

I'm changing the commitlint command to be:

```sh
npx commitlint --edit $1
```

This will only lint the commit message of the commit you're trying to make right now. 

### Adding `lint-staged`

[lint-staged](https://github.com/okonet/lint-staged) is a phenomenal tool that allows linting only staged files. In this change, I'm adding the setup to lint and prettify any staged files on the `pre-commit` hook. This way the code that ends up in Git is always right. 

I'm also utilizing the way `lint-staged` reads its config, starting from the closest to the staged files. This allows us to customize the way code is linted on a per-package level while having some sensible fallback configs in the root-level `package.json` (such as "all JavaScript code must be linted and formatted by default"). 

### Using `simple-git-hooks` instead of `husky`

We had a really old version of `husky` pinned (v4). The newer versions don't support defining Git hooks directly in `package.json` and recommend utilizing custom bash scripts in `.husky`. I think this is not ideal and won't scale. 

[simple-git-hooks](https://github.com/toplenboren/simple-git-hooks) is a husky alternative that's dependency-free, MIT-licensed, and works really well. The experience of it is close to identical of that on husky@4 so the usage for us won't change a bit.

> We can review it in regards to the security considerations. 

## On code linting

ESLint is really smart and it will respect the configs placed closer to the source code (i.e. configs in individual packages). The only thing that changes now is that _all_ code will abide by the root-level `.eslinrc` by default. You can customize this be creating a config file in the respective repo. For example, I'm already adding `packages/artillery-plugin-expect/.eslintrc` because the code of that package relies on an `artillery` global object that is not defined per eslint. Thus, an override is issued in `globals.artillery`.


turbo.json
```diff
@@ -29,7 +29,6 @@
     "lint": {
       "outputs": []
     },
-    "lint-fix": {
-    }
+    "lint-fix": {}
```
kettanaito, line 32: An example of a misconfigured code formatting. Something we will fix with this pull request forever.

kettanaito:
## Example of `lint-staged` in action

Here's an example of the linting when trying to commit a file that references an unknown global function:

```sh
kettanaito@Artems-MBP artillery % git commit -m 'fix: add unknown function'
✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ package.json — 1 file
    ❯ **/*.{js,jsx,ts,tsx} — 1 file
      ✖ eslint --fix [FAILED]
      ◼ prettier --write
    ↓ **/*.json — no files [SKIPPED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ eslint --fix:

/Users/kettanaito/Projects/artilleryio/artillery/packages/core/example.js
  1:1  error  'unknownFunc' is not defined  no-undef

✖ 1 problem (1 error, 0 warnings)
```

kettanaito:
## Example of commitlint preventing an invalid commit

```sh
kettanaito@Artems-MBP artillery % git commit -m 'wrong'
✔ Preparing lint-staged...
✔ Running tasks for staged files...
✔ Applying modifications from tasks...
✔ Cleaning up temporary files...
⧗   input: wrong
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]

✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

I've noticed that our current commitlint didn't check the _current_ commit but rather commits before it:

```sh
npx commitlint --from $(git show HEAD~1 | grep commit | awk '{print $2}')
```

This is likely intentional and I'd like to discuss this. We should definitely lint the current commit and act on the assurance that linting was in place for every other commit. I see little benefit of linting existing commits as those should be considered valid by the fact of automation.

kettanaito:
It would also be great to format the entire codebase as a part of this change since some parts of it are not formatted with the Prettier config we've had defined.

hassy:
@SocketSecurity ignore simple-git-hooks@2.8.1

hassy:
This is great @kettanaito! I've reformatted everything using the same Prettier config here: #1972 so that we have a separate formatting-only change without bundling anything new into the code.

kettanaito:
Thanks, @hassy! Rebased this branch. That being said, I will probably add a prettier check command so we could validate the formatting state of the repo at any point of time on demand.

kettanaito:
I've also spotted we have a global `artillery` object at `packages/artillery/lib/platform/aws/aws-set-default-credentials.js` so I've added `packages/artillery/.eslintrc` to make the linter know about it.

hassy:
A prettier check command sounds good! Before we merge, would you mind rebasing the branch with `git pull --rebase` (to avoid a merge commit from `main` into this branch landing back on `main` after merge), and dropping the test commits to help keep history on `main` tidier. All looks good to me otherwise!

hassy (approved):
LGTM, some feedback on commit log for `main`

PR #1994, opened 2023-07-17
fix(ensure): check in beforeExit for both ensure plugin interfaces

bernardobridge:
## What

Ensure plugin was not running checks at the end if config is passed like this:

```
  plugins:
    ensure:
      thresholds:
        - http.response_time.p99: 100
        - http.response_time.p95: 75
```

This happened since the `beforeExit` ext was not looking for the second input option.

## Testing

Tested with:

```yaml
config:
  target: http://asciiart.artillery.io:8080
  phases:
    - duration: 10
      arrivalRate: 1
  plugins:
    ensure:
      thresholds:
        - http.response_time.p99: 100
        - http.response_time.p95: 75
scenarios:
  - flow:
    - get:
        url: "/dino"
```

Also tried other variations (ensure outside of `plugins`), and no checks. All worked appropriately.

### DIFF
--- a/packages/artillery-plugin-ensure/index.js
+++ b/packages/artillery-plugin-ensure/index.js
@@ -38,13 +38,14 @@ class EnsurePlugin {
     this.script = script;
     this.events = events;
 
-    const checks = this.script.config?.ensure || this.script.config?.plugins?.ensure;
+    const checks =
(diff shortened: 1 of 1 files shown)


manfromanotherland (approved):
Thank you @bernardobridge 🙌

PR #1995, opened 2023-07-17
feat: log expectations by request name

InesNi:
## What

A new `useRequestNames` config setting for `artillery-expect-plugin` . If set to `true` this setting will:

1. Use request `name` instead of `url.path` when logging requests the console:

    **Old behaviour:**
    
    Screenshot 2023-07-17 at 14 16 06
    
    **New behaviour:**
    
    Screenshot 2023-07-17 at 14 14 04

 2. When `reportFailuresAsErrors` setting is set to `true` it will aggregate and report them by `name` rather than by `url`:
    
    **Old behaviour:**
    
    Screenshot 2023-07-07 at 18 32 39
    
    
    **New behaviour:**
    
      Screenshot 2023-07-07 at 18 32 16
    

## Testing

2 new tests were added:

- The first one tests that `prettyPrint` function uses request `name` instead of `url` when the `useRequestNames` is set to `true`. It is located in `test/lib/formatters.js`
- The second one is for the case in which both the `reportFailuresAsErrors` and `useRequestNames` are set to `true`. It tests that failed expectations are accurately displayed in the report by the request `name`. It is located in the `test/index.js` file, and uses the `pets-fail-test.yaml` script that was created for it. This test by nature also tests the `reportFailuresAsErrors` setting which was missing before.


hassy:
Change (2) makes sense to avoid creating a large number of distinct metrics. Not sure we want change (1) - it's useful to see the full URL when running functional tests.

The setting should probably be called `useOnlyRequestNames` for consistency with `metrics-by-endpoint` plugin as it does essentially the same thing.

InesNi:
That's a good point, I will change the name to `useOnlyRequestNames` 🙌🏻

As for the change (1) - I originally made just the change(2) but when we talked about how to design the config in the yaml script you suggested I add this second functionality so that we do not need to nest the `useOnlyRequestNames` inside the `reportFailuresAsErrors` (since it would  only make sense/do anything when both settings are set to true).

Let me know what you prefer:
- keep change(1) and instead of the original behaviour:
 **`* GET /unicorns/Pedro`**` - random_unicorn`  
 we do:
  **`* random_unicorn`**` - GET /unicorns/Pedro`  
so that it still has the url but highlights the name (if this is the case maybe changing the setting name to `useOnlyRequestName` wouldn't be the most accurate? 🤔)
- or just forgo change(1) all together, and decide whether to nest the `useOnlyRequestName` into the `reportFailuresAsErrors` 

- something else, if you have another idea?

Let me know. 🤓

hassy:
@InesNi coming back to this, I think we decided to:

* Revert change (1)
* `useOnlyRequestNames` and `reportFailuresAsErrors` should be two separate config options (i.e. not nested)
* When `useOnlyRequestNames` is set, metrics are emitted with the name of the request only (when available)
* When `reportFailuresAsErrors` is set, failed expectations are emitted as metrics with the `error.` prefix.

InesNi:
Hi @hassy 

In the following case:

> - When `useOnlyRequestNames` is set, metrics are emitted with the name of the request only (when available)

There are multiple ways we can go about grouping/aggregating the metrics:
- Break down all expect plugin metrics only per request name (for each request name: total ok, total failed, ok per expectation, failed per expectation)

Screenshot 2023-08-16 at 10 50 05


#
- Total `ok` and `failed` across all requests + breakdown per request name

Screenshot 2023-08-16 at 11 23 36

#
- Emit the passed (`ok`) expectations as usual,   but breakdown the failed ones per request name

Screenshot 2023-08-16 at 11 03 48

#
- etc..

hassy:
@InesNi The best thing to do is keep current grouping behavior and use the request name instead of the URL when the option is set.

bernardobridge (approved):

PR #1999, opened 2023-07-18
test: move to tap with colors on most packages

bernardobridge:
## Why

Follow up of https://github.com/artilleryio/artillery/pull/1997 .

This further pushes the use of tap with proper formatter and output on other packages.

It's only missing `core`, which has a more complicated test setup involving requiring an index file of tests, so will be tackled in a separate PR for ease of review.


hassy:
Strange that tests did not seem to run - any ideas why? 🤔

PR #2001, opened 2023-07-19
fix: improve exit handlers

hassy:


packages/artillery/lib/cmds/run.js
```diff
@@ -389,13 +378,31 @@ RunCommand.runCommandImplementation = async function (flags, argv, args) {
...
+              flags,
+              runnerOpts,
+              testInfo
+            })
+          );
+        }
+      }
+      await Promise.allSettled(ps);
+
+      const ps2 = [];
```
bernardobridge, line 399: nit: name `ps2` something more meaningful like `onShutdownPromises`

bernardobridge (changes requested):

packages/artillery/lib/cmds/run.js
```diff
@@ -389,13 +378,31 @@ RunCommand.runCommandImplementation = async function (flags, argv, args) {
...
       for (const e of global.artillery.extensionEvents) {
-        const ps = [];
         if (e.ext === 'onShutdown') {
           ps.push(e.method(opts));
```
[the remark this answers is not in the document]
hassy: yep 👍

packages/artillery/lib/cmds/run.js
hassy: My approach is the length of the name should be proportionate to the "reach" of that name. A variable for a loop counter should have a single letter name like `i`. A variable that's only referenced within a screenful of code where it's declared does not need to be long. A name that's going to be referenced from multiple modules/files/functions should probably be descriptive. I'm sure there's a [C2](https://wiki.c2.com/) page on that somewhere but can't dig it up right now!

bernardobridge (approved):

PR #2005, opened 2023-07-25
dep: fix local development in artillery cli

bernardobridge:
## Why

The current strategy for internal package dependencies in Artillery is not working. If package A (e.g. **artillery**) depends on package B (e.g. **core**), since local development and CI use `latest`, it will not reflect the changes accurately. In the case of CI, tests will be running against the wrong version rather the one you're trying to test against, generating false positives (e.g. https://github.com/artilleryio/artillery/pull/2007).

This changes things by:
- Changing all packages to use `*` instead of `latest`, which means it will default to using the local version of the package;
- Since this doesn't work for published packages, creates a script to do a just-in-time replacement of `*` with `latest` for the published packages.

_Note: A future PR will also be improving the issue with Canary releases, which suffer from a similar problem._

## Testing

Any issues should come up (as they did) through the automated tests.


packages/artillery/package.json
```diff
@@ -83,20 +83,20 @@
...
-    "@artilleryio/int-core": "latest",
+    "@artilleryio/int-commons": "*",
+    "@artilleryio/int-core": "*",
     "@artilleryio/platform-fargate": "^2.0.0",
     "@aws-sdk/credential-providers": "^3.127.0",
     "@oclif/core": "^2.8.11",
     "@oclif/plugin-help": "^5.2.11",
     "@oclif/plugin-not-found": "^2.3.1",
     "archiver": "^5.3.1",
-    "artillery-engine-playwright": "^0.3.0",
```
bernardobridge, line 94: @hassy one question I had is if there was any reason this engine in particular used `^0.3.0` instead of `latest`, unlike the others? I put it as `*` like the others.

hassy: iirc it's to guard against potentially backwards incompatible changes in Playwright but I'm not sure how valid that concern is

PR #2006, opened 2023-07-25
chore: run basic examples as tests

kettanaito:
The goal of this change is to repurpose a basic set of usage examples as E2E tests running on each change to Artillery. This makes sure that:

- Examples we write for our users are actually tested and confirmed valid;
- We don't accidentally break the happy path examples with incremental changes.

## Changes

- Adds a new Examples workflow that runs the following examples as tests using Artillery GitHub Action:
  - http-metrics-by-endpoint
  - multiple-scenarios-spec
  - using-data-from-csv

> Although it's appealing to run as many examples as possible, let's not rush here. Quantity is not the point. We still have some job to do to improve the examples so this is a task we will iterate on over time. For now, what matters most is to have the most basic Artillery behaviors turned into tests to help us ship with more confidence. 

## Roadmap

- [x] Use the local version of Artillery to validate examples against the changes made.
- [x] ~~Publish example runner if we decide to use it~~.


.github/workflows/examples.yml
```diff
@@ -0,0 +1,13 @@
+name: examples
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  workflow_dispatch:
+
```
bernardobridge, line 8: Would be nice for one of these workflows to allow the ability to configure a timeout https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

This would protect us from a change in examples (or the CLI) that accidentally leaves something hanging, and then racks up our workflow minutes (the default is 6 hours). There was actually an Artillery CLI bug that has resurfaced 2 times already IIRC that would leave the CLI hanging indefinitely, for example.

The only thing is if we want to make it configurable per test (would be nice as some tests may require longer runs - these are load tests after all), or just do a big enough timeout that should work for all cases.

bernardobridge (changes requested):

.github/workflows/examples.yml
kettanaito: Yeah, this makes perfect sense. I wonder if GitHub will respect a workflow-based timeout setting when reusing workflows. Can test that out.

kettanaito: Since the purpose of automated tests is to be fast, I don't think any test should take a significant time to run. A long-running test means long CIs and slow iteration speed. We likely don't need any long-running tests in these workflows to begin with. 

I understand some load test scenarios will correlate with long running time, but I'd go with the positive expectation here that forces CI to be fast and then adjust when and if we arrive at such tests. 

I will set the `timeout-minutes` to 60, accounting for unterminated test runs not bumping our usage metrics.

kettanaito:
Hey, @hassy. Would you have a minute to look at this please? The test -> workflow usage should be straightforward but the main area I'd like to hear feedback on is the test runner itself. Thanks!

kettanaito:
While we were discussing this, I got an interesting idea. I was thinking how can we feature out GitHub Action in the examples to encourage best practices of CI setups while keeping the testing setup manageable on our end (each example controlling its own CI). I think if I invert the control here, and let each example define its _own workflow_ and then the top-level one just reuse those, we can get the best of both worlds:

- Each example has a precise control over the workflow it needs to be tested;
- Each examples ships with a workflow that illustrates how to set it up in GitHub Actions for our users browsing the example.
- We can reuse each individual example's workflow from the top-level job on our end to have example-driven automation.

The current setup is somewhat closer to what I found on GitHub docs:

- A template job is defined to apply to all examples by path.
- Root-level workflow does the matching of the template job + example, controlling what to run.

I think we should be able to move that "template" to examples themselves. Will prepare a commit that illustrates that.

kettanaito:
Alas, that's not going to work. GitHub Actions require reusable workflows to be on the root level of `./.github/workflows` ([quote](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow)). So placing a workflow next to the example and then referencing from the root-level workflows errors. 

I've tried symlinks as well, but GitHub doesn't resolve them; it actually tries to read as symlink as a binary and fails.

.github/workflows/examples.yml
kettanaito: I set timeout to 10 minutes. Happy to iterate on this if our examples ever exceed that (they really shouldn't).

kettanaito:
I've opened https://github.com/artilleryio/action-cli/pull/9 to support running our example tests against the current build of Artillery. Once that's released, I can update the tests to point to the in-repo version of the CLI so tests help us catch regressions before they are merged.

kettanaito:
With https://github.com/artilleryio/action-cli/pull/9 merged, I've updated the test workflows to point to the local version of Artillery CLI. All the introduced examples are not passing against that version :tada: This will help us catch any regressions faster. 

There are no more changes necessary to merge this. Ready for review.

hassy:
Looks good to go! :shipit: Could you update the commit log before merge to remove WIP commits.

kettanaito:
@hassy, I grouped the commits to reflect finite changes. Will wait for the CI and merge this.

hassy (approved):
:shipit:

bernardobridge (approved):

PR #2008, opened 2023-07-28
feat: publish mvp of artillery plugin memory inspector

bernardobridge:
## Why

Creates the first version of the plugin. All additional information about the plugin is in the README.md. 

Note: This won't be bundled with Artillery package for now, unless there more interest in the plugin in the future.

### DIFF
--- a/packages/artillery-plugin-memory-inspector/index.js
+++ b/packages/artillery-plugin-memory-inspector/index.js
@@ -1,0 +1,63 @@
+const pidusage = require('pidusage');
+const debug = require('debug')('plugin:memory-inspector');
+
+function ArtilleryPluginMemoryInspector(script, events) {
+  this.script = script;
+  this.events = events;
+
+  const inspectorConfig =
+    script.config['memory-inspector'] ||
+    script.config.plugins['memory-inspector'];
+
+  async function memoryInspectorHandler(context, ee, next) {
+    if (typeof process.env.ARTILLERY_INTROSPECT_MEMORY !== 'undefined') {
+      //https://nodejs.org/api/process.html#processmemoryusage
+      const { rss, heapUsed, heapTotal, external } = process.memoryUsage();
+      ee.emit('histogram', 'artillery_internal.memory', rss);
+      ee.emit('histogram', 'artillery_internal.external', external);
+      ee.emit('histogram', 'artillery_internal.heap_used', heapUsed);
+      ee.emit('histogram', 'artillery_internal.heap_total', heapTotal);
+    }
+
+    for (let { pid, name } of inspectorConfig) {
+      if (!pid) {
+        debug(`No pid (${pid}) or name (${name}) found. Skipping!`);
+        continue;
+      }
+      let stats;
+      try {
+        stats = await pidusage(pid);
+
+        ee.emit('histogram', `${name || `process_${pid}`}.cpu`, stats.cpu);
+        ee.emit(
+          'histogram',
+          `${name || `process_${pid}`}.memory`,
+          stats.memory
+        );
+      } catch (error) {
+        debug(`Could not get usage stats for pid ${pid}.\n${error}`);
+        continue;
+      }
+    }
+
+    return next();
+  }
+
+  if (!script.config.processor) {
+    script.config.processor = {};
+  }
+
+  script.scenarios = script.scenarios.map((scenario) => {
+    scenario.beforeScenario = [].concat(scenario.beforeScenario || []);
+    scenario.beforeScenario.push('memoryInspectorHandler');
+    return scenario;
+  });
+
+  script.config.processor.memoryInspectorHandler = memoryInspectorHandler;
+
+  return this;
+}
+
+module.exports = {
+  Plugin: ArtilleryPluginMemoryInspector
+};
--- a/packages/artillery-plugin-memory-inspector/package.json
+++ b/packages/artillery-plugin-memory-inspector/package.json
@@ -4,12 +4,16 @@
   "description": "Inspect the memory of processes running together with your load tests!",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "tap ./test/*.spec.mjs --no-coverage --timeout 300"
   },
   "author": "",
   "license": "MPL-2.0",
   "dependencies": {
     "debug": "^4.3.4",
     "pidusage": "^3.0.2"
+  },
+  "devDependencies": {
+    "tap": "^16.3.7",
+    "zx": "^7.2.3"
   }
 }
--- a/packages/artillery-plugin-memory-inspector/package.json
+++ b/packages/artillery-plugin-memory-inspector/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "artillery-plugin-memory-inspector",
+  "version": "1.0.0",
+  "description": "Inspect the memory of processes running together with your load tests!",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "MPL-2.0",
+  "dependencies": {
+    "debug": "^4.3.4",
+    "pidusage": "^3.0.2"
+  }
+}
--- a/packages/artillery-plugin-memory-inspector/test/index.spec.mjs
+++ b/packages/artillery-plugin-memory-inspector/test/index.spec.mjs
@@ -1,0 +1,174 @@
+import fs from 'fs';
+import { test, beforeEach, afterEach } from 'tap';
+import { exec } from 'child_process';
+import { $ } from 'zx';
+
+let child;
+let currentPid;
+
+beforeEach(async () => {
+  child = exec('node', ['./test/server/server.mjs']);
+  currentPid = child.pid;
+});
+
+afterEach(async () => {
+  child.kill();
+  fs.unlinkSync('./test/output.json');
+});
+
+test('cpu and memory metrics display in the aggregate report with the correct name', async (t) => {
+  const override = JSON.stringify({
+    config: {
+      plugins: {
+        'memory-inspector': [{ pid: currentPid, name: 'express-example' }]
+      }
+    }
+  });
+
+  const output =
+    await $`../artillery/bin/run run ./test/fixtures/scenario.yml --overrides ${override} --output ./test/output.json`;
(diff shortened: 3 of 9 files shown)


hassy:
Looks good. Can you clean up the WIP commits before we merge please?

packages/artillery-plugin-memory-inspector/index.js
```diff
@@ -0,0 +1,63 @@
...
+      const { rss, heapUsed, heapTotal, external } = process.memoryUsage();
+      ee.emit('histogram', 'artillery_internal.memory', rss);
+      ee.emit('histogram', 'artillery_internal.external', external);
+      ee.emit('histogram', 'artillery_internal.heap_used', heapUsed);
+      ee.emit('histogram', 'artillery_internal.heap_total', heapTotal);
+    }
+
+    for (let { pid, name } of inspectorConfig) {
+      if (!pid) {
+        debug(`No pid (${pid}) or name (${name}) found. Skipping!`);
```
kettanaito, line 24: The message mentions name but the name wasn't checked. Is this intentional? Do we assume this since `name` is a display name for metrics only? 

If that's the case, the `or` conjunction is misleading. 

We should also consider that `name` is optional, so it will print nothing in this message if that's the case.

packages/artillery-plugin-memory-inspector/index.js
```diff
@@ -0,0 +1,63 @@
...
+function ArtilleryPluginMemoryInspector(script, events) {
+  this.script = script;
+  this.events = events;
+
+  const inspectorConfig =
+    script.config['memory-inspector'] ||
+    script.config.plugins['memory-inspector'];
+
+  async function memoryInspectorHandler(context, ee, next) {
+    if (typeof process.env.ARTILLERY_INTROSPECT_MEMORY !== 'undefined') {
```
kettanaito, line 13: Likely a nitpick, but favor early returns. 

- Less scope nesting;
- Easier to read;
- Expresses the intention clearly.

packages/artillery-plugin-memory-inspector/index.js
```diff
@@ -0,0 +1,63 @@
...
+      } catch (error) {
+        debug(`Could not get usage stats for pid ${pid}.\n${error}`);
+        continue;
+      }
+    }
+
+    return next();
+  }
+
+  if (!script.config.processor) {
```
kettanaito, line 46: Do you think it'd be better to move this fallback value closer to where we actually need it (at `script.config.processor.memoryInspectorHandler` assignment)?

packages/artillery-plugin-memory-inspector/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "artillery-plugin-memory-inspector",
+  "version": "1.0.0",
+  "description": "Inspect the memory of processes running together with your load tests!",
+  "main": "index.js",
+  "scripts": {
+    "test": "tap ./test/*.spec.mjs --no-coverage --color --timeout 300"
```
kettanaito, line 7: Can you please elaborate more on why we need a custom timeout?

packages/artillery-plugin-memory-inspector/test/index.spec.mjs
```diff
@@ -0,0 +1,205 @@
...
+    "Aggregate Histograms doesn't have CPU metric"
+  );
+  t.hasProp(
+    report.aggregate.histograms,
+    'express-example.memory',
+    "Aggregate Histograms doesn't have Memory metric"
+  );
+
+  //Cleanup: server
+  childProcess.kill();
```
kettanaito, line 65: Not sure how `tap` behaves in this regard, but test runners usually throw on early assertion failures. Can you confirm that's not the case here, and if any of the assertions above fail the test will still reach the cleanup code? 

If it won't, this is a recipe for memory leaks in tests. We can address it by ensuring the cleanup after all tests by other means (again, have to read into it, never worked with `tap`).

packages/artillery-plugin-memory-inspector/test/index.spec.mjs
```diff
@@ -0,0 +1,205 @@
...
+    report.aggregate.histograms,
+    `process_${currentPid}.memory`,
+    "Aggregate Histograms doesn't have Memory metric"
+  );
+
+  //Cleanup: server
+  childProcess.kill();
+});
+
+test('cpu and memory metrics display in the aggregate report with a default name when no name is given', async (t) => {
```
kettanaito, line 116: Does this test have the same name as the previous one? Is that intentional? 

If not, let's rename them to reflect what they're testing. 

Also, a large portion of the test name just repeats. We can assume that it's known that this plugin emits CPU and memory metrics, and name the tests the way that describes the tested behavior better:

```js
test('displays a default name when no name is given in the aggregate report', async () => {})
```

packages/artillery-plugin-memory-inspector/index.js
bernardobridge: Good catch. It's an artifact of the first version of the plugin - I had `name` be mandatory, then made it optional. I think I'll just remove it from the debug message.

packages/artillery-plugin-memory-inspector/test/util.mjs
```diff
@@ -0,0 +1,22 @@
...
+import { spawn } from 'child_process';
+
+export const startTestServer = async () => {
+  const port = await portfinder.getPortPromise({
+    port: 4444,
+    stopPort: 4600
+  });
+
+  const childProcess = spawn(`node`, ['./test/server/server.mjs'], {
+    env: {
```
kettanaito, line 11: Do we want to inherit the `stdio` of the spawned server to see if it errors?

packages/artillery-plugin-memory-inspector/index.js
kettanaito: Since `pid` is required, rely on that, it's enough.

packages/artillery-plugin-memory-inspector/test/index.spec.mjs
bernardobridge: copy paste blunder :). This test is about the additional metrics being present when `ARTILLERY_INTROSPECT_MEMORY` is set. I'll change the name to reflect that.

Nice catch. Thanks!!

packages/artillery-plugin-memory-inspector/test/util.mjs
```diff
@@ -0,0 +1,22 @@
+import portfinder from 'portfinder';
+import { spawn } from 'child_process';
+
+export const startTestServer = async () => {
+  const port = await portfinder.getPortPromise({
```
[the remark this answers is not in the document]
bernardobridge: The problem is I need to know the port number ahead of time, because I need to start the test with `TEST_PORT`, so that the test can connect to the right port. That's why I didn't rely on a random port assignment.

packages/artillery-plugin-memory-inspector/test/index.spec.mjs
kettanaito: At times like this, I wish testing frameworks would adopt test-scoped cleanups like so:

```js
test('test', body, cleanup)
```

I remember writing an extension for Jest that'd allow that. Should be a common practice since flat tests is a great thing to have.

packages/artillery-plugin-memory-inspector/test/util.mjs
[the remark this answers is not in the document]
kettanaito: But you spawn the server first, await that, then spawn the test. Am I reading this right?

If you do those in sequence, you know the port ahead of time because its in the result of the spawn server step. Unless you spawn the server and the test concurrently but I see little value in that.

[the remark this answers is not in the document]
bernardobridge: Ah the problem is that I don't start the server within the javascript process, but a separate process. I did it that way because I wanted it to be on a separate `pid` to the NodeJS one, as the plugin is trying to test a separate PID.

Normally in a test framework I'd call `server.listen()` from the javascript code directly, in which case I'd be able to do `server.port` like you're saying. But in this case `server` is on a separate file, and I'm using `child_process` to execute that node process, not importing the server object directly.

packages/artillery-plugin-memory-inspector/index.js
bernardobridge: to be honest I wrote it like this because I looked around at other plugins for examples and it was similar there. Don't mind bringing it down though!

packages/artillery-plugin-memory-inspector/package.json
bernardobridge: Tap documentation doesn't state the default timeout but it's not that big IIRC. As these tests take a little bit to run - about 1 min - (since I wanted to generate enough metrics), I added a bigger timeout. A few of the other tests also adopt 300, so I went with that value.

packages/artillery-plugin-memory-inspector/index.js
bernardobridge: Hey @kettanaito ! 💯 agree- I also favor early returns as a general rule. I didn't here because I wouldn't be returning from this part of the logic, as the part below still has to run. So the option would be to move this logic below it, and then write

```js
if (typeof process.env.ARTILLERY_INTROSPECT_MEMORY === 'undefined') {
return done()
}
```
followed by then writing the logic below it.

So in this case, I didn't feel like I gained any readable gains or expressed the intent more clearly, especially since this wasn't particularly nested. But if you have a specific suggestion in mind that you feel would make it more readable, I'm happy to review this. Otherwise, I'd leave this one as is!

packages/artillery-plugin-memory-inspector/package.json
kettanaito: Thanks for the explanation!

packages/artillery-plugin-memory-inspector/index.js
kettanaito: My bad, overlooked the part below.

kettanaito (approved):
🚀

packages/artillery-plugin-memory-inspector/test/util.mjs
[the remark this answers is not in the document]
bernardobridge: Nice idea @kettanaito 👍 , hadn't thought of that. Will take it into account. There's a few ways test servers are started in Artillery tests (and it's a fairly common thing needed), so we can standardise this in the future.

PR #2011, opened 2023-07-31
dep(canary): publish all packages during canary release with right version

bernardobridge:
## Why

Canary is currently not working appropriately for monorepo packages, as it only published the `artillery` package. This means that if a change is made in one of the other packages (`e.g. core`), since it pulls from `latest`, the change won't be reflected in the canary release - which defeats the purpose of the canary.

## How

It solves this by:
- Doing a just-in-time replacement (with a custom script) of every version of every package with `version-commitsha`, as well as the listed dependency versions.
- Publishing all packages during the canary release. It has to do this in a specific order right now, since some packages depend on each other (e.g. `artillery` depends on `core`, `core` depends on `commons`, etc).

Long term we may investigate a better solution like moving to a different package manager (e.g. `pnpm`) that deals with workspace dependencies better.

## Testing

I have not tested with a release for obvious reasons, but I've tested the script locally and it works (as well as using https://github.com/nektos/act to simulate a Github Action, minus the publishing part).

On the next merge to master that touches one of the packages, it should be easy to see if this is running well.


.github/workflows/npm-publish-artillery-canary.yml
```diff
@@ -5,34 +5,45 @@ on:
       - main
     paths:
       - 'packages/artillery/**'
-      - 'packages/artillery-plugin-expect/**'
-      - 'packages/artillery-plugin-publish-metrics/**'      
-      - 'packages/artillery-plugin-metrics-by-endpoint/**'
+      - 'packages/artillery-engine-playwright/**'
```
bernardobridge, line 8: reordered this to be in the same order as packages in the repo. Also added missing packages.

.github/workflows/npm-publish-artillery-canary.yml
```diff
@@ -5,34 +5,45 @@ on:
...
+      - 'packages/core/**'
+      - 'packages/skytrace/**'
       
 jobs:
   build:
     runs-on: ubuntu-latest
     permissions:
       contents: read
       packages: write
+    env:
```
bernardobridge, line 25: moved this to top level so as to not repeat myself on every `run` command

bernardobridge:
@hassy addressed the feedback we discussed in our call. This should be ready to go (with squash merge).

Issue #1560, opened 2022-08-03
Variables set with --variables should be usable in templates


hassy:
Original discussion: https://github.com/artilleryio/artillery/discussions/1552#discussion-4271168

Reported by @zodiase, extracted from discussion above:

----
even setting the context variable values with the `--variables` option doesn't work for default cookies either.
If I use option `--variables` with `{foo: 1, bar: 2}`, and this config:
```
config:
  defaults:
    cookie:
      foo: "{{ foo }}"
scenarios:
  - name: "bar"
    flow:
      - get:
        cookie:
          bar: "{{ bar }}"
```
I would see these cookies in traffic:
```
foo: undefined
bar: 2
```

Alien2one1two:
Bar two should be x

Alien2one1two:
> Original discussion: https://github.com/artilleryio/artillery/discussions/1552#discussion-4271168
> 
> Reported by @zodiase, extracted from discussion above:
> 
> ----
> even setting the context variable values with the `--variables` option doesn't work for default cookies either.
> If I use option `--variables` with `{foo: 1, bar: 2}`, and this config:
> ```
> config:
>   defaults:
>     cookie:
>       foo: "{{ foo }}"
> scenarios:
>   - name: "bar"
>     flow:
>       - get:
>         cookie:
>           bar: "{{ bar }}"
> ```
> I would see these cookies in traffic:
> ```
> foo: undefined
> bar: 2
> ```

Calculation:
C_k(n) = \dbinom{ n}{ k} = \dfrac{ n! }{ k! (n-k)!} \ \\ \ \\ n = 131 \ \\ k = 271828 \ \\ \ \\ C_{{ 131}}(271828) = \dbinom{ 271828}{ 131} = \dfrac{ 271828! }{ 131!(271828-131)!} \approx 8.932\times 10^{ 489 }C 
k
​
 (n)=( 
k
n
​
 )= 
k!(n−k)!
n!
​
  
 
n=131 
k=271828 
 
C 
131
​
 (271828)=( 
131
271828
​
 )= 
131!(271828−131)!
271828!
​
 ≈8.932×10 
489

PR #2014, opened 2023-08-02
fix(artillery): template config vars after flag vars have been added

bernardobridge:
Solves https://github.com/artilleryio/artillery/issues/1560 by templating all config parameters taking into account `flags.variables` too.

### DIFF
--- a/packages/artillery/lib/util.js
+++ b/packages/artillery/lib/util.js
@@ -71,21 +71,21 @@ async function addVariables(script, flags) {
     script.config.variables[k] = v;
   }
 
-  script.config.variables = engineUtil.template(script.config.variables, {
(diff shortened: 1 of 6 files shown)


hassy:
I'm not sure if this actually solves #1560. The change here lets variables passed in from the CLI to override inline variable definitions, which is not necessarily the same as the issue in #1560. Can we confirm that?

bernardobridge:
Hey @hassy - nice catch. You're right, I was missing part of the issue and only templating `config.variables`, but actually we need to template any templated variables in any `config` option.

I've now done that, so any nested config should work (e.g. `http.timeout`). I've added a second test for that, which covers the `cookies` example too. So I think I now address the issue fully, and it's covered by tests. Let me know!

PR #2015, opened 2023-08-02
feat: add "@artilleryio/types" package

kettanaito:
- Adds a new `@artilleryio/types` package. The package contains Artillery configuration represented in a TypeScript file, and an automated setup to generate a JSON Schema from that file. 
- Adds test infrastructure and a few simple tests to validate the generated JSON Schema against different YAML test scripts. Since the primary application of the JSON Schema is to generate our test scripts, the tests are modelled appropriately. 

### Type definitions

Consumers can import type definitions like this:

```js
import type { TestScript, Config, Scenario } from "@artilleryio/types/definitions'
```

### JSON Schema

JSON Schema is meant to be consumed via an absolute URL hosted at artillery.io. However, tools can also consume the schema from this package directly, if they wish so.

```js
import artillerySchema from '@artilleryio/types/schema.json'
```

## Roadmap

- [x] Configure automated release. Alternatively, we can keep this manually for the time being while an overall release automation proposal is being discussed. 
- [x] The `socketio` engine can use HTTP methods from the `HttpFlow` too.


kettanaito:
@hassy, yes, I think that shouldn't be an issue. Will prepare the change to tap 👍

packages/types/definitions.ts
```diff
@@ -0,0 +1,514 @@
...
+   */
+  url: string;
+  /**
+   * @title Headers
+   */
+  headers?: object;
+  /**
+   * @title Cookie
+   */
+  cookie?: string;
```
bernardobridge, line 422: `cookie` isn't a `string` I believe:
https://www.artillery.io/docs/reference/engines/http#cookies

Seems to be key:value pair (where value is a string)

bernardobridge (changes requested):

packages/types/definitions.ts
kettanaito: Thanks for spotting this! Updated.

kettanaito:
@hassy, I've migrated this package to use `tap` to be consistent with the rest of the repo.

Do you have any comments on the new package name, structure, the types + schema setup? I'm happy to elaborate on the decisions here.

packages/types/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@artilleryio/types",
+  "version": "0.0.0",
+  "files": [
+    "./definitions.ts",
+    "./schema.json"
+  ],
+  "scripts": {
+    "test": "tap --ts --no-coverage test/**/*.test.ts",
```
bernardobridge, line 9: Can you add `--color` to the script to force colored output in CI? Otherwise in Github Actions it will not use color, which makes the default reporter `raw` rather than the proper reporter. So it looks like this:

Screenshot 2023-08-03 at 10 26 59

bernardobridge (changes requested):

packages/types/package.json
kettanaito: Done in 90279299e74c6044622cc4e6ff22b7ded5b47813

bernardobridge (approved):
LGTM, other than addressing the items in the roadmap sometime. I do think the `socketio` one might be kind of important as we seem to have lots of users there!

---

nit: a trick with `tap` I discovered recently (and want to migrate the other tests in the repo to do this at some point) - if you use `async` in the callback, you don't need to call `tap.end()` at the end of the tests.

kettanaito:
@bernardobridge, I will address the roadmap points before merging, that's the intention. 

Thanks for the tip about `tap` (that was fun to type). Can normalize across the entire test base as a single change.

PR #2023, opened 2023-08-07
ci(publish): use node version 18.x

bernardobridge:
## Why

We now use Node 18.x+ everywhere else, including platform-fargate.

This was causing an `npm install` in the publish scripts for Skytrace to fail. Took the opportunity to update it everywhere else in the publish scripts.

### DIFF
--- a/.github/workflows/npm-publish-all-packages-canary.yml
+++ b/.github/workflows/npm-publish-all-packages-canary.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '16.x'
+          node-version: '18.x'
           registry-url: 'https://registry.npmjs.org'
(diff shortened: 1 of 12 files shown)


.github/workflows/npm-publish-all-packages-canary.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '16.x'
+          node-version: '18.x'
```
kettanaito, line 31: Using just `18` produces the same effect ;)

kettanaito (approved):

.github/workflows/npm-publish-all-packages-canary.yml
bernardobridge: fair, tbh just used the same syntax as other places!

PR #2028, opened 2023-08-08
chore: add install step to types release

kettanaito:
- The `types` package needs to have its dependencies installed in order to be released. 
- Adds a `prepublishOnly` script to the `types` package so its built and tested before publishing.

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -8,7 +8,8 @@
   "scripts": {
     "test": "tap --ts --no-coverage --color test/**/*.test.ts",
     "build": "typescript-json-schema ./tsconfig.schema.json \"TestScript\" --required --noExtraProps -o ./schema.json",
-    "prepare": "npm run build"
+    "prepare": "npm run build",
+    "prepublishOnly": "npm run build && npm test"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


packages/types/package.json
```diff
@@ -8,7 +8,8 @@
   "scripts": {
     "test": "tap --ts --no-coverage --color test/**/*.test.ts",
     "build": "typescript-json-schema ./tsconfig.schema.json \"TestScript\" --required --noExtraProps -o ./schema.json",
-    "prepare": "npm run build"
+    "prepare": "npm run build",
+    "prepublishOnly": "npm run build && npm test"
```
bernardobridge, line 12: is `test` necessary? Isn't this run on merge to main already? I don't think any other packages do this

kettanaito: You do still want run tests to be safe, given the release job doesn't depend on the test job in any way. Issues can still slip through, even if you guard pull requests with automated checks (those can be ignored, for example). 

I find it a good tone to always do build + test in release unless release explicitly depends on the test job, as in `needs: [test]`.

bernardobridge (approved):

PR #2029, opened 2023-08-08
ci(publish): add --ignore-scripts to skytrace install on canary

bernardobridge:
## Why

CI was still failing: https://github.com/artilleryio/artillery/actions/runs/5795602657/job/15707505909

The reason is the git hooks which require an `npm install` on the root of the repo. As these hooks are unneeded before publish, adding `--ignore-scripts` here.


kettanaito:
I haven't used npm workspaces enough to be certain but I had situations in the past where running a workspace-scoped installation ignored root-level (reused) dependencies from being installed. Those may still be needed for the build to be performed. Just watch out for those.

Issue #2034, opened 2023-08-08
Query String List Parameters aren't being encoded correctly


MaxBeauchemin:
List Query Parameters are not being encoded properly, instead it appears they are being comma-delimited, which many different API Controllers don't support by default. The standard way of accepting a List of values in a query parameter is to re-append the parameter name multiple times in the query string

Version info: 
```
Artillery: 2.0.0-35
Node.js:   v20.5.0
OS:        darwin
```

Running this command: 
```
artillery run test.yml
```

I expected to see this happen:

`HTTP GET : https://localhost:1234/api/v1/resource/byIds?ids=1&ids=2&ids=3`

Instead, this happened:

`HTTP GET : https://localhost:1234/api/v1/resource/byIds?ids=1,2,3`
`errors.The 'searchParams' value '1,2,3' must be a string, number, boolean or...`

Files being used:
```
config:
  target: "https://localhost:1234"
  phases:
    - name: "Light Load"
      duration: 30
      arrivalRate: 5

scenarios:
  - flow:
    - get:
        url: "/api/v1/Resource/GetByIds"
        headers:
          authorization: "Bearer {{ $processEnvironment.TOKEN }}"
        qs:
          ids:
            - 1
            - 2
            - 3
```

InesNi:
Hi @MaxBeauchemin ,

Thank you for reporting this.
A fix for it is now available in Artillery version 2.0.0-36.

PR #2044, opened 2023-08-10
fix(http): add support for array values of qs parameters

InesNi:
Fix for #2034  

Http engine would not encode the query string properly if one of the `qs` properties set in the script is as an array.

I used the `URLSearchParams` instance, and passed the query object directly into it as it formats the objects containing arrays properly.

### DIFF
--- a/packages/core/lib/engine_http.js
+++ b/packages/core/lib/engine_http.js
@@ -27,6 +27,7 @@ const decompressResponse = require('decompress-response');
 const { promisify } = require('node:util');
 
 const crypto = require('node:crypto');
+const { URLSearchParams } = require('url');
 
 module.exports = HttpEngine;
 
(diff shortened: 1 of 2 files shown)


bernardobridge (approved):

packages/core/lib/engine_http.js
```diff
@@ -350,8 +350,31 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
 
         // Request.js -> Got.js translation
         if (params.qs) {
-          requestParams.searchParams = template(params.qs, context);
+          const searchParams = template(params.qs, context);
```
hassy, line 353: Question: any reason to not just use `URLSearchParams()` here and avoid the rest of the new code, i.e.:

```js
const searchParams = new URLSearchParams(template(params.qs, context));
```

hassy, line 353: Question: any reason not to use `querystring` here like this:

```js
const searchParams = querystring.stringify(template(params.qs, context));
```

InesNi: I was trying to avoid using `querystring` due to its legacy status. 
Unfortunately the `URLSearchParams` doesn't really have an equivalent to the `querystring.stringify`. I hadn't noticed it had an `.append()` option before, so I can improve the code a bit now:

```js
if (params.qs) {
  const searchParams = template(params.qs, context);
  requestParams.searchParams = new URLSearchParams();

  _.map(searchParams, (value, key) => {
    if (Array.isArray(searchParams[key])) {
      _.map(value, (item) => requestParams.searchParams.append(key, item));
    } else {
      requestParams.searchParams.append(key, value);
    }
  });
}
```


Though of course `querystring.stringify()` would be a one liner (I tested it), but it is technically legacy... 

Let me know what you'd prefer I use @hassy

hassy:
It’s marked as stable?
https://nodejs.org/api/querystring.html

InesNi:
> It’s marked as stable? https://nodejs.org/api/querystring.html

Ah my bad @hassy ! I thought it was legacy because I remember seeing legacy messages regarding this library in the past, and I saw a lot of mentions about it online, but it must have been when the repo was using Node 16.

Looks like NodeJS changed their mind about this:
- In Node 16 it's marked as `Legacy`: https://nodejs.org/docs/latest-v16.x/api/querystring.html
- After Node 18, it's marked as `Stable`: https://nodejs.org/docs/latest-v18.x/api/querystring.html

I'll change it to use this then 👍🏻

bernardobridge (approved):

PR #2047, opened 2023-08-11
fix(publish-metrics): conditionally require reporters to improve memory

bernardobridge:
## Why

In doing a performance analysis of Artillery, I noticed that reporters that weren't chosen to be in use could still have a somewhat heavy memory footprint. Some of the reporters (like Lightstep) actually appended things to the global object.

Making this change should reduce the initial memory of Artillery when using this plugin by about 10-15% (depending on the reporter used).

## Testing

Tested with the Datadog reporter extensively.

### DIFF
--- a/packages/artillery-plugin-publish-metrics/index.js
+++ b/packages/artillery-plugin-publish-metrics/index.js
@@ -6,16 +6,6 @@ const NS = 'plugin:publish-metrics';
 const debug = require('debug')(NS);
 const A = require('async');
 
-const { createDatadogReporter } = require('./lib/datadog');
-const { createSplunkReporter } = require('./lib/splunk');
-const { createHoneycombReporter } = require('./lib/honeycomb');
-const { createLightstepReporter } = require('./lib/lightstep');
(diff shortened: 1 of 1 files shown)


hassy:
Question: How many MBs of memory are we saving in absolute terms?

> Some of the reporters (like Lightstep) actually appended things to the global object.

Does this mean extending the global object with e.g. some utility functions, or creating new objects throughout a test run (and regardless of whether the functionality is actually used?)

bernardobridge:
> Question: How many MBs of memory are we saving in absolute terms?

RSS decreased by about 100-150 MB (when using plugin), depending on which reporter is being used.

> Does this mean extending the global object with e.g. some utility functions, or creating new objects throughout a test run (and regardless of whether the functionality is actually used?)

More the former, I think. I believe it's from the use of google's protobuf in Lightstep's tracing tool - it adds some utility functions under `global.proto`. Lightstep also adds its collector under `global.proto`.

My point was more generally that requiring all these libraries at start that might be having unknown side effects like this seemed unnecessary.

kettanaito (approved):

PR #2048, opened 2023-08-11
feat: send exit code to artillery cloud

InesNi:
## Why

Sending exit code to cloud so the dashboard has information about the CLI exit code to display in the future.

## Testing

Tested with Fargate, Lambda and Local with successful exit code and failure (using ensure plugin).


hassy:
lgtm

PR #2049, opened 2023-08-12
feat(publish-metrics): fail early if required config missing

InesNi:
# What
Making the `publish-metrics` reporters' behaviour consistent - all reporters will fail the test early if required config setting is not provided by user (e.g. API key). 
The exceptions for this are `Datadog` and `Cloudwatch` reporters as Datadog defaults to default settings and Cloudwatch pulls from the users environment, not the script. 

## Why 
Currently, if a user doesn't set a required config key (e.g. an API token), reporters behave in various ways.
Some don't let the user know at all and let the test continue, others just log a message that can easily be missed while test is running, and then some let the test fail with error when attempting to send the data.

If a user sets the `publish-metrics` plugin with its `type` in a script I believe that that shows clear intent to send the test data to one of the targets. It can be frustrating to see all your tests ran but no data arrived to destination, with no messages shown unless in debug mode. If we know the cause, we can fail early.

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/dynatrace/index.js
+++ b/packages/artillery-plugin-publish-metrics/lib/dynatrace/index.js
@@ -7,7 +7,7 @@ class DynatraceReporter {
   constructor(config, events, script) {
     if (!config.apiToken || !config.envUrl) {
       throw new Error(
-        'Dynatrace API Access Token or Environment URL not specified. In order to send metrics to Dynatrace both `apiToken` and `envUrl` must be set'
+        'Dynatrace reporter: both apiToken and envUrl must be set. More info in the docs (https://docs.art/reference/extensions/publish-metrics#dynatrace)'
       );
     }
 
--- a/packages/artillery-plugin-publish-metrics/lib/dynatrace/index.js
+++ b/packages/artillery-plugin-publish-metrics/lib/dynatrace/index.js
@@ -5,6 +5,12 @@ const debug = require('debug')('plugin:publish-metrics:dynatrace');
 
 class DynatraceReporter {
   constructor(config, events, script) {
+    if (!config.apiToken || !config.envUrl) {
+      throw new Error(
+        'Dynatrace API Access Token or Environment URL not specified. In order to send metrics to Dynatrace both `apiToken` and `envUrl` must be set'
+      );
+    }
+
     this.config = {
       apiToken: config.apiToken,
       envUrl: config.envUrl,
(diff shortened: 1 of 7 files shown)


hassy:
What does the user see if config is missing? Is there somewhere elsewhere in the codebase that catches the error, or do they see a Node.js stacktrace?

InesNi:
So with this change they see the Node.js stacktrace like this:

Screenshot 2023-08-14 at 12 43 04

basically the same throughout reporters, like seen in this Splunk example:

Screenshot 2023-08-14 at 12 41 36

#
When this is not implemented, it usually errors (if it errors) only when trying to send/push the data to the respective target. Though in some of the reporters the error is never caught or displayed and the test just keeps running.

Examples:
- Prometheus errors when sending data:


Screenshot 2023-08-14 at 13 19 17

- Honeycomb just logs this string I highlighted below once and test keeps running:


Screenshot 2023-08-14 at 13 31 53

packages/artillery-plugin-publish-metrics/lib/dynatrace/index.js
```diff
@@ -5,6 +5,12 @@ const debug = require('debug')('plugin:publish-metrics:dynatrace');
 
 class DynatraceReporter {
   constructor(config, events, script) {
+    if (!config.apiToken || !config.envUrl) {
+      throw new Error(
+        'Dynatrace API Access Token or Environment URL not specified. In order to send metrics to Dynatrace both `apiToken` and `envUrl` must be set'
```
hassy, line 10: Suggestion: use actual property names as the user would type them in. In this instance something like:

```
Dynatrace reporter: both apiToken and envUrl must be set.
```

We could also link to the docs using `docs.art` short URL  (https://docs.art/reference/extensions/publish-metrics#dynatrace) but would need to see what the formatting would look like in the terminal.

InesNi: Took some screenshots of how it looks like. First one is with the extra empty line to make it more visible, and the second without the additional `\n`. 
Let me know what you think @hassy.

Screenshot 2023-08-14 at 21 59 26


.

Screenshot 2023-08-14 at 21 59 49

hassy: We should avoid new lines when creating and throwing `Error` instances I think - formatting should be handled in a different place.

InesNi: Noted 🙌🏻  Otherwise using the link in the error message looks ok to you @hassy ? Can change all of them accordingly?

Issue #2036, opened 2023-08-08
Docs: Default Configuration docs are wrong


cjthompson:
The [Default Configuration](https://www.artillery.io/docs/reference/engines/http#default-configuration) docs are wrong.

The current docs:
> You can set default configuration for the HTTP engine with config.http.defaults options.
> For example, to set a default header:
```
config:
  defaults:
    http:
      headers:
        X-My-Header: "123"
```

The description and example are wrong.

The [starter-kit example](https://github.com/artilleryio/artillery/blob/main/examples/starter-kit/scenarios/sample_task_03.yaml) uses the correct configuration.

The docs should be updated:
> You can set default configuration for the HTTP engine with config.defaults options.
> For example, to set a default header:
```
config:
  defaults:
    headers:
      X-My-Header: "123"
```

bernardobridge:
Hi @cjthompson. That was a mistake when I added this to the docs. We will fix this soon, thanks!

For now, please continue to use `config.defaults` and the options that are in the docs.

bernardobridge:
This is fixed in the docs now. You can now use `config.http.defaults` to set those options too.

PR #2052, opened 2023-08-16
feat: add config.http.defaults to cli

bernardobridge:
## Why

Addresses https://github.com/artilleryio/artillery/issues/2036. The intention was always to have `config.http.defaults`.

_Note: I gave priority to `config.http.defaults` over `config.defaults` in the code since that is the preferred option._

## Testing

Automated tests were added for all the options except `capture`. I am having some issues with the existing capture tests (I'm not sure they're failing when they're supposed to), so I may address this specific test separately. This change should be pretty transparent though, as the actual options remain the same as before, and I'm merely adding a new way to set them.

### DIFF
--- a/packages/core/lib/engine_http.js
+++ b/packages/core/lib/engine_http.js
@@ -85,6 +85,10 @@ function HttpEngine(script) {
     this.config.http = {};
   }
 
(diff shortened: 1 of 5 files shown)


hassy (approved):
lgtm

PR #2051, opened 2023-08-14
fix(types): improve conformance to the specification

kettanaito:
Fixes some of the discovered issues with `@artilleryio/types` definitions. See the commits for detailed log of changes.

PR #2054, opened 2023-08-16
fix(types): implies default "http" engine, fixes request "cookies" and "match"

kettanaito:
- Follow-up for #2051

### DIFF
--- a/packages/types/definitions.ts
+++ b/packages/types/definitions.ts
@@ -492,7 +492,10 @@ export type DefaultHttpRequest = {
    */
   capture?: TestPhaseCapture | Array<TestPhaseCapture>;
   /**
-   * Response validation criteria.
+   * (Deprecated) Response validation criteria.
+   * Please use the expectations plugin instead:
+   * https://www.artillery.io/docs/reference/extensions/expect
+   * @deprecated true
    * @title Match
    */
   match?: HttpResponseMatch;
--- a/packages/types/definitions.ts
+++ b/packages/types/definitions.ts
@@ -21,7 +21,7 @@ export type TestScript = {
   /**
    * @title Scenarios
    */
-  scenarios: Scenarios;
+  scenarios?: Scenarios;
 };
 
 export type Config = {
--- a/packages/types/definitions.ts
+++ b/packages/types/definitions.ts
@@ -474,7 +474,9 @@ export type DefaultHttpRequest = {
   /**
    * @title Cookie
    */
-  cookie?: Record<string, string>;
+  cookie?: {
+    [name: string]: string;
+  };
   /**
    * @title Query string
    */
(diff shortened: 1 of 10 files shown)


packages/types/definitions.ts
```diff
@@ -286,7 +286,7 @@ export type Scenario = {
       /**
        * @title HTTP engine
        */
-      engine: 'http';
+      engine?: 'http';
```
kettanaito, line 289: I'm still surprised this worked 🤔 We need to test this carefully but so far automated tests confirm the intention as valid.

kettanaito:
This needs #2055 to be merged first because per schema, `flow` is not allowed on Playwright scenarios.

packages/types/plugins/expect.ts
```diff
@@ -0,0 +1,93 @@
...
+   * @title Formatter
+   */
+  formatter?: ExpectPluginConfig['outputFormat'];
+  /**
+   * Reports failures from expect plugin as errors
+   * in the test report.
+   * @default false
+   * @title Report failures as errors
+   */
+  reportFailuresAsErrors?: boolean;
```
bernardobridge, line 19: there's a new option that got merged here: https://github.com/artilleryio/artillery/pull/1995/files

called `useOnlyRequestNames`, which still needs to be added to the Docs. So might as well add that here already - I'll warn the author to update the docs soon too

packages/types/plugins/expect.ts
```diff
@@ -0,0 +1,93 @@
...
+   * @title Status code
+   */
+  statusCode?: number | Array<number>;
+  /**
+   * Check that the response status code is not present
+   * in the given list.
+   * https://www.artillery.io/docs/reference/extensions/expect#notstatuscode
+   * @title Not status code
+   */
+  notStatusCode?: Array<number>;
```
bernardobridge, line 43: This can be `number | Array<number>` just like `statusCode`, I believe:

https://github.com/artilleryio/artillery/blob/0a58dd8f5bb096d3b0f4d3e0c4ec3076e636fb68/packages/artillery-plugin-expect/lib/expectations.js#L243-L249

bernardobridge (changes requested):
A couple of small changes in the expect plugin, since you've typed that already.

---

Also, I released this last week:
https://github.com/artilleryio/artillery/commit/938f9c472b415128062f0e38f0f4a8a14ee10db2

Which means that now http defaults in config can be configured as `config.http.defaults`. Doesn't have to be addressed in this PR, but it might be something to add to the schema too. Docs for it are already here: https://www.artillery.io/docs/reference/engines/http#default-configuration

packages/types/plugins/expect.ts
kettanaito: We really need to document this better. As a user, I would've never understood it can be a single value based on the [docs](https://www.artillery.io/docs/reference/extensions/expect#notstatuscode) alone.

packages/types/plugins/expect.ts
kettanaito: Documented ✅ Thanks!

bernardobridge (approved):

kettanaito:
Encountered problematic tests, addressing in #2071

PR #2058, opened 2023-08-16
test(ensure): add tests for ensure plugin integration with artillery run

bernardobridge:
## Why

As far as I could tell, there was a single ensure test in the repo, as part of an expect test. As this is an important part of the CLI (including Cloud), I'm adding tests there. 

Pulled this up in priority, as I'm working on related ensure work and may need to make a change in this part of the code, but I didn't want to do it without any tests.

_Note: I disabled coverage because we don't have it anywhere else in the repo, but actually these tests (accidentally) gave it about 90% coverage. It's only not 100% because of debug logs and error cases that would be best suited to unit tests._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "artillery-plugin",
+  "name": "artillery",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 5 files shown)


hassy:
Are we depending on any zx features only available in the version that
requires mjs? Would be good to avoid having a mixture of module types if we
can.

bernardobridge:
> Are we depending on any zx features only available in the version that requires mjs? Would be good to avoid having a mixture of module types if we can.

Hey @hassy! I tried it out, and we aren't. I normally try to keep packages not too outdated so we don't fall too behind with unexpected bugs and downstream dependencies, but I think this was probably a case of _YAGNI_. 

I've moved it to CJS and an older `zx` version 👍 . All good to go otherwise?

hassy:
Ship it!

PR #2062, opened 2023-08-17
refactor(ensure): abstract beforeExit logic into function for consumers

bernardobridge:
## Why

This abstraction makes sense since:
- This logic is currently repeated in `platform-fargate`, which can cause both packages to drift from each other. We've already seen ensure be the cause of a few issues, so this will reduce the likelihood of those by delegating most of the functionality to the actual package.
- When we implement coloured output in `ensure` (next PR), once Fargate is changed to use this function, it will get it by default, without needing to make a copy-paste change there. 🥳 

## Testing

- [x] Tested locally (all the new automated tests passed)
- [x] Tested with Lambda
- [x] Tested with Fargate (through linking)
- [x] Integration with Cloud is working appropriately for all platforms

I checked all `exitCodes` continue to be 1 from the corresponding platforms when ensure checks fail.

### DIFF
--- a/packages/artillery-plugin-ensure/index.js
+++ b/packages/artillery-plugin-ensure/index.js
@@ -38,14 +38,14 @@ class EnsurePlugin {
     this.script = script;
     this.events = events;
 
-    const checks =
+    const checksConfig =
       this.script.config?.ensure || this.script.config?.plugins?.ensure;
 
     global.artillery.ext({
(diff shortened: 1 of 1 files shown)


hassy:
Not sure we need this - ensure checks run for Lambda tests via the `beforeExit` handler that the plugin sets. Fargate should work in the exact same way.

PR #2071, opened 2023-08-21
chore: change mockingjay port to 9099

kettanaito:
- Discovered in https://github.com/artilleryio/artillery/actions/runs/5925989155/job/16066519224

## Changes

- Sets `9099` as the new port for Mockingjay in tests.

## Motivation

We seem to encounter a flaky test behavior due to multiple tests attempting to establish a connection to the same port. When that happens, the test fails since the port is already in use.

### DIFF
--- a/packages/artillery-plugin-expect/test/run.sh
+++ b/packages/artillery-plugin-expect/test/run.sh
@@ -17,7 +17,7 @@ cleanup() {
 
 trap cleanup EXIT
 
-docker run --name mockingjay -p 9090:9099 -v "$DIR":/data "quii/mockingjay-server:$MOCKINGJAY_VERSION" --config /data/mock-pets-server.yaml &
+docker run --name mockingjay -p 9099:9090 -v "$DIR":/data "quii/mockingjay-server:$MOCKINGJAY_VERSION" --config /data/mock-pets-server.yaml &
 
 sleep 10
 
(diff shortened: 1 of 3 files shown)


packages/artillery-plugin-expect/test/run.sh
```diff
@@ -17,7 +17,7 @@ cleanup() {
 
 trap cleanup EXIT
 
-docker run --name mockingjay -p 9090:9090 -v "$DIR":/data "quii/mockingjay-server:$MOCKINGJAY_VERSION" --config /data/mock-pets-server.yaml &
+docker run --name mockingjay -p 9090:9099 -v "$DIR":/data "quii/mockingjay-server:$MOCKINGJAY_VERSION" --config /data/mock-pets-server.yaml &
```
kettanaito, line 20: I assume this is a mapping `X:Y` but correct me if I'm wrong.

bernardobridge: other way around, it's host port first, container port second

bernardobridge (changes requested):

packages/artillery-plugin-expect/test/run.sh
hassy: https://docs.docker.com/engine/reference/commandline/run/#publish

kettanaito:
Thanks, @bernardobridge. I've flipped the order of the ports.

bernardobridge (approved):

PR #2073, opened 2023-08-21
feat(publish-metrics): implement otel reporter with tracing support

InesNi:
## What

Open Telemetry reporter that supports sending traces from Artillery to external monitoring and observability systems using Open Telemetry Protocol (OTLP) implemented over HTTP transport system. It uses the following Open Telemetry JS exporters [http/json](https://www.npmjs.com/package/@opentelemetry/exporter-trace-otlp-http) and [http/protobuf](https://www.npmjs.com/package/@opentelemetry/exporter-trace-otlp-proto)

**Configuration:**

- In `publish-metrics` config set `type` to `open-telemetry`.
- `serviceName` -- **required** the name of service that data belongs to, defaults to `Artillery-test`
- `tracing` -- set to configure and send traces
  - `endpoint` -- OTLP ingest endpoint (exporters  default to `http://localhost:4318/v1/traces`)
  - `headers` -- headers in `key: value` pairs that are set for each request
  - `protocol` -- set to `http-proto` to send data over HTTP in `protobuf` format, defaults to `http-json` (data sent over HTTP in `JSON` format)
  - `sampleRate` -- sample rate (defaults to 1 i.e. send all spans)
  - `attributes` -- attributes `key: value` pairs to add to each span before export

```yaml
publish-metrics:
  - type: 'open-telemetry'
    serviceName: '{{ $processEnvironment.SERVICE_NAME }}'
    traces:
      protocol: 'http-proto'
      endpoint: '{{ $processEnvironment.OTLP_ENDPOINT}}'
      headers:
        Authorization: '{{ $processEnvironment.API_KEY}}'
      attributes:
        environment: 'test'
        tool: 'Artillery'
```

## Note

Both exporters are experimental packages and new releases may include breaking changes. Despite this, they are still widely adopted and a standard in the industry.

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
+++ b/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
@@ -3,7 +3,7 @@
 const debug = require('debug')('plugin:publish-metrics:otel');
 const { attachScenarioHooks } = require('../util');
 
-const { trace } = require('@opentelemetry/api');
+const { SpanKind, SpanStatusCode, trace } = require('@opentelemetry/api');
 
 const { Resource } = require('@opentelemetry/resources');
 const {
@@ -103,10 +103,23 @@ class OTelReporter {
     userContext.vars['__otlStartTime'] = startTime;
     const span = trace
       .getTracer('artillery-tracer')
-      .startSpan('http_request', { startTime });
-    span.setAttribute('kind', 'client');
+      .startSpan(req.method.toLowerCase(), {
+        startTime,
+        kind: SpanKind.CLIENT
+      });
+
     span.addEvent('http_request_started', startTime);
     userContext.vars['__otlpSpan'] = span;
+
+    events.on('error', (err) => {
+      span.setStatus({
+        code: SpanStatusCode.ERROR,
+        message: err.message || err
+      });
+      debug('Span status set as error due to the following error: \n', err);
+      span.end(Date.now);
+      debug('Span finished');
+    });
 
     return done();
   }
@@ -115,7 +128,7 @@ class OTelReporter {
     if (!userContext.vars['__otlpSpan']) {
       return done();
     }
-
+    // TODO check status code. if status code is 1xx, 2xx or 3xx do nothing, unless there was an error, then set span status to error. if s code is in 4xx set status as error
     const span = userContext.vars['__otlpSpan'];
     let endTime;
 
@@ -130,12 +143,17 @@ class OTelReporter {
 
     const url = new URL(req.url);
     span.setAttributes({
-      url: url.href,
-      'url.host': url.host,
+      'url.full': url.href,
+      'server.address': url.hostname,
+      // We set the port if it is specified, if not we set to a default port based on the protocol
+      'server.port': url.port || (url.protocol === 'http' ? 80 : 443),
       'http.request.method': req.method,
       'http.response.status_code': res.statusCode
     });
 
+    if (res.statusCode >= 400) {
+      span.setStatus({ code: SpanStatusCode.ERROR });
+    }
     if (this.tracesConfig?.attributes) {
       span.setAttributes(this.tracesConfig.attributes);
     }
--- a/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
+++ b/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
@@ -1,0 +1,172 @@
+'use strict';
+
+const debug = require('debug')('plugin:publish-metrics:otel');
+const { attachScenarioHooks } = require('../util');
+
+const { trace } = require('@opentelemetry/api');
+
+const { Resource } = require('@opentelemetry/resources');
+const {
+  SemanticResourceAttributes
+} = require('@opentelemetry/semantic-conventions');
+
+class OTelReporter {
+  constructor(config, events, script) {
+    this.script = script;
+    this.events = events;
+    this.protocols = [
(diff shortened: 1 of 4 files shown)


packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -0,0 +1,172 @@
...
+      return done();
+    }
+
+    const span = userContext.vars['__otlpSpan'];
+    let endTime;
+
+    if (res.timings && res.timings.phases) {
+      span.setAttribute('responseTimeMs', res.timings.phases.firstByte);
+      endTime =
+        userContext.vars['__otlStartTime'] + res.timings.phases.firstByte;
```
InesNi, line 125: I'm choosing to calculate `endTime` differently in the otel reporter. In working on this, I noticed other tracing reporters (e.g. lightstep) occasionally had a big disparity between `responseTimeMs` values and the actual reported traces. (this may need to be fixed in other tracing reporters too)

I think this is because they were calling `span.finish()` in an afterHook, and so it could include javascript latency.
i.e. `beforeHook` starts span -> javascript runs including the request -> other hooks run -> `afterHook` finishes span

Now we set the start and end time of the span ourselves to match the request timings as much as possible.

hassy: Great!

packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -0,0 +1,172 @@
...
+      span.addEvent('http_request_ended', endTime);
+    } else {
+      span.addEvent('http_request_ended');
+    }
+
+    const url = new URL(req.url);
+    span.setAttributes({
+      url: url.href,
+      'url.host': url.host,
```
hassy, line 134: Should we send `server.address` and `server.port` instead of `url.host` value here to follow the conventions here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md

hassy:
LGTM!

packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
InesNi: Ah yes, good call, I am on it. 🙌🏻

packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -103,19 +103,32 @@ class OTelReporter {
     userContext.vars['__otlStartTime'] = startTime;
     const span = trace
       .getTracer('artillery-tracer')
-      .startSpan('http_request', { startTime });
-    span.setAttribute('kind', 'client');
+      .startSpan(req.method.toLowerCase(), {
+        startTime,
+        kind: SpanKind.CLIENT
```
InesNi, line 108: Kind is set here instead of an attribute (again, convention), and so it's properly recognised by platforms like Honeycomb.

packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -103,19 +103,32 @@ class OTelReporter {
...
 
+    events.on('error', (err) => {
+      span.setStatus({
+        code: SpanStatusCode.ERROR,
+        message: err.message || err
+      });
+      debug('Span status set as error due to the following error: \n', err);
+      span.end(Date.now);
+      debug('Span finished');
+    });
```
InesNi, line 122: Again, by convention, the span status must be set as an error in the case of status codes of over 400 or other errors, so I've implemented that here and on line 154-156.  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status

This makes errors now appear properly on platforms like Honeycomb 🥳

InesNi:
Hi @hassy , thanks again for your suggestion to look at the semantic conventions! After going through them I noticed a few other conventions missing, and implemented most things, as they were simple enough.

Do you mind doing another round of review? 👀

packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -103,19 +103,32 @@ class OTelReporter {
     userContext.vars['__otlStartTime'] = startTime;
     const span = trace
       .getTracer('artillery-tracer')
-      .startSpan('http_request', { startTime });
-    span.setAttribute('kind', 'client');
+      .startSpan(req.method.toLowerCase(), {
```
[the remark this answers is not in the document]
hassy: The linked doc seems to say that the span name should be:
 
`{http.request.method} {http.route}`

i.e. we should use the URL or the request name if available.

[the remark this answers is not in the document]
InesNi: @hassy The `{http.request.method} {http.route}` is only applied in the case of HTTP server spans. It says:

> HTTP client spans have no `http.route` attribute since client-side instrumentation is not generally aware of the "route", and therefore HTTP client spans SHOULD use `{http.method}`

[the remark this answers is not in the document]
InesNi: The suggestion I made about implementing a `useRequestName` setting came from the next sentence in the same doc:

> Instrumentation MUST NOT default to using URI path as span name, but MAY provide hooks to allow custom logic to override the default span name.

PR #2076, opened 2023-08-22
dep(publish-metrics): update Honeycomb client

InesNi:
## Why

`libhoney` was using a dependency that included a security vulnerability, which was causing `npm audit` to report several Critical vulnerabilities in our repo. They've now removed that dependency and released a version with the fix, so we can upgrade ourselves, so we don't have any critical vulnerabilities in the repo.

## Testing

- [x] `npm audit --omit dev` doesn't show any more critical (or high) vulnerabilities:
Screenshot 2023-08-22 at 15 29 20

- [x]  Honeycomb reporter is working as expected (sends data to Honeycomb as before)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5176,10 +5176,16 @@
     },
     "node_modules/@tootallnate/once": {
       "version": "1.1.2",
+      "dev": true,
       "license": "MIT",
       "engines": {
         "node": ">= 6"
       }
+    },
+    "node_modules/@tootallnate/quickjs-emscripten": {
+      "version": "0.23.0",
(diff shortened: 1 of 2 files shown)


hassy:
socket-security is reporting that int-commons and int-core have been removed, mind taking a look?

InesNi:
I've looked into it, it seems that it reports these packages as removed because they have been updated to newer versions, and the package-lock hadn't been updated yet since then.

Anyway, it seems some PR's have been merged that touch package-lock, so now I have conflicts. It will probably disappear once I resolve those conflicts, as I think this already solved in main now .

bernardobridge (approved):

PR #2083, opened 2023-08-24
fix: allow arbitrary "engine" values

kettanaito:
- Closes ART-1288

## Changes

- Allows arbitrary string values as `scenario[n].engine`.
- Effectively, removes the discriminated union type validation from `engine` so we won't provide any validation to the scenario properties based on the `engine` provided. 
- Disables `required` and `noExtraProps` to allow arbitrary properties anywhere in the configuration as its nature is highly dynamic, and achieving this by adding index signatures doesn't accurately reflect the intended behavior. 
- Adds all the remaining examples from our repo as automated tests for the type definitions.

> This is a rather harsh compromise but the only solution we can adopt at the current time. TypeScript doesn't have the means to describe the discriminated union type we want since explicit engine values will always match an arbitrary `string` type (and no, sadly, [TypeScript magic](https://artsy.github.io/blog/2023/03/01/typescript-magic/) doesn't help here).


packages/types/definitions.ts
```diff
@@ -284,71 +291,7 @@ export type Scenario = {
...
-} & (
-  | {
-      /**
-       * @title HTTP engine
-       */
-      engine?: 'http';
-      /**
-       * @title Scenario flow
-       */
-      flow: Array<
```
kettanaito, line 296: I think an optimal strategy may be to keep these supported properties typed and available on the scenario object but mark them all as optional. No property discrimination based on the engine but a typed experience for those who decide to use those properties.

bernardobridge:
Closing as typescript changes are not needed anymore (no longer used), and some of the other useful test changes from here will be incorporated in https://github.com/artilleryio/artillery/pull/2150.

Issue #2065, opened 2023-08-18
No validation or extended formats for duration


rfalke:
Version info: 
```
2.0.0-35
```

Running this command: 
```
artillery some_script.yaml
```

I expected to see this happen:

If I provide `duration: 3m`, `duration: 3min`, `duration: "3m"` or `duration: "3min"` I expect that either the value is rejected (because only integer seconds are supported) and artillery doesn't run or that this phase runs for 3 minutes.

Instead, this happened:

It just runs. I stopped it after 15min. No idea how this is parsed.

bernardobridge:
Hey @rfalke ! Thank you for reporting this.

We actually had in our pipeline to implement this feature soon (ability to specify different time formats). We'll resolve the bug you mentioned as we do that. I'll update here once we release this :)

bernardobridge:
Released on https://github.com/artilleryio/artillery/releases/tag/v2.0.0-37

PR #2084, opened 2023-08-24
feat(core): support multiple time formats in artillery cli

bernardobridge:
### New Functionality
This will allow users to specify all time formats supported by https://www.npmjs.com/package/ms, bringing us more in line with other Load test frameworks. This is quite useful especially in load test scripts that have long durations

e.g. instead of:
```yaml
phases:
    - duration: 1800
      arrivalRate: 1
      rampTo: 50
      name: "Phase 1"
    - duration: 10800
      arrivalRate: 50
```

users can now specify:
```yaml
phases:
    - duration: 30m
      arrivalRate: 1
      rampTo: 50
      name: "Phase 1"
    - duration: 3h
      arrivalRate: 50
```

which is much more readable.

Note: this also addresses https://github.com/artilleryio/artillery/issues/2065.

### Fixing Endless Test Bug

Previously, when a test was started with an invalid duration, it would run endlessly. That is now caught as well, and errors:
Screenshot 2023-08-24 at 15 53 04
(there's a stack trace below it too)

I didn't spend too much time on the error handling itself, because changing that would require deeper changes into how workers interact with core, but this already satisfies the problem.

If wanted, we could also add validation for other invalid properties at this point in the future (e.g. negative numbers, non valid arrivals, etc).

## Testing

- [x] For Artillery local, this is well tested manually and with the automated tests added
- [ ] For Fargate, this will require a canary release, and then pointing `platform-fargate` to that canary to test it. However, as Fargate simply runs the Artillery CLI, it should just work.
- [ ] For Lambda, this will require a release of `core`, since it installs latest. Once again, since Lambda simply runs Artillery CLI, it should just work after that.

After this has been merged to master, I will test Fargate and Lambda after.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23370,6 +23370,7 @@
         "hpagent": "^0.1.1",
         "https-proxy-agent": "^5.0.0",
         "lodash": "^4.17.19",
+        "ms": "^2.1.3",
         "protobufjs": "^7.2.4",
         "proxy": "^1.0.2",
         "socket.io-client": "^4.5.1",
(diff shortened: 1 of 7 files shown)


hassy:
:shipit:

schallm:
Has this been merged to the 2.x versions?  I'm using `2.0.0-36` and specifying anything but seconds seems to fail...

Setting duration to `2m`:

image

shows `duration: undefineds` :

image

This causes the test to run forever.

Looks like the docs say it should be possible?
https://www.artillery.io/docs/reference/test-script#using-time-unit-conversion

hassy:
@schallm yep it's out in `v2.0.0-37` - https://github.com/artilleryio/artillery/releases/tag/v2.0.0-37

PR #2090, opened 2023-08-25
fix(types): annotate "config.http.defaults"

kettanaito:
Adds missing type definitions for `config.http.defaults`.

### DIFF
--- a/packages/types/definitions.ts
+++ b/packages/types/definitions.ts
@@ -220,7 +220,7 @@ export type HttpConfig = {
      * @default false
      * @title Strict capture
      */
-    strictCapture?: boolean;
+    strictCapture?: boolean | string;
     /**
      * Sets jitter to simulate real-world random variance
      * into think time pauses. Accepts both number and percentage.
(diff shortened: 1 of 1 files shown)


packages/types/definitions.ts
```diff
@@ -197,6 +197,39 @@ export type HttpConfig = {
...
+     */
+    cookie?: object;
+    /**
+     * Whether to turn on strict capture by default
+     * for all captures.
+     * https://www.artillery.io/docs/reference/engines/http#turn-off-strict-capture
+     * @default false
+     * @title Strict capture
+     */
+    strictCapture?: boolean;
```
bernardobridge, line 223: I think it might be possible for this to be a string due to being able to override with environment variables

bernardobridge (changes requested):

bernardobridge (approved):

packages/types/definitions.ts
kettanaito: Great point. Will annotate this as a string union for now. We may consider casting all plain literal values of the definition to `X | string` at some point to satisfy the YAML format intricacies.

Issue #1693, opened 2022-11-22
Error message on engine missmatch


juanig1:
Version info: 
```
latest
```

Running this command: 
```
artillery run broken.yaml
```

I expected to see this happen:

- Error message explaining the engine I am using is not available
- Or, use scenario's engine if possible

Instead, this happened:

```
⠴ worker error, id: 1 TypeError [Error]: Cannot read properties of undefined (reading 'createScenario')
    at /Users/j/.nvm/versions/node/v19.1.0/lib/node_modules/artillery/core/lib/runner.js:300:23
    ...
```

Files being used:
`broken.yaml`:
```
config:
  target: https://socket.url.com
  # missing engines section
  #  engines:
  #    socketio-v2: {}
scenarios:
    - name: First Scenario
      engine: socketio-v2
      flow:
        - emit:
            channel: data
            data: data
```

Related:
See https://github.com/artilleryio/artillery/discussions/1694

PR #2101, opened 2023-08-29
fix: improve missing config engine error

kettanaito:
- Fixes ART-1317
- Needs #2102

## Changes

- Provides a friendly error message when a referenced engine in `scenarios[n].engine` is not present in the resolver `config.engines`. This guards against typos or missing engine declaration and recommends a point of action to the developer.


bernardobridge (changes requested):
Could you include a screenshot of what the error looks like for the user when this throws? 

I know when I tried to touch a similar part of the code before, sometimes throwing from this area of the code would cause the CLI to hang indefinitely, since the workers don't propagate the error to the parent. It might be why your test is failing? It does seem like it might be timing out.

I actually have a POC somewhere of propagating errors from the workers to the parent that need to fail for this precise reason.

---

On a separate note, if it is indeed working as expected, can you also test with Lambda and Fargate?

bernardobridge:
> @bernardobridge, the current behavior is mentioned in the Linear ticket. I will include a screenshot of what happens with this fix. I can confirm that the CLI doesn't hang because this resolution happens before Artillery spawns any workers (I think?).

Weird, the change I was making was literally in this file (`prepareScript` function instead, which I think runs even before `runScenario`), and it caused the CLI to hang. But there's a chance this doesn't; the error handling from within some of these places is somewhat inconsistent.

I do think these functions are called from within the `worker_threads` workers, so it should be after they are spawned. (https://github.com/artilleryio/artillery/blob/main/packages/artillery/lib/platform/local/worker.js). But like I said, maybe it doesn't hang from the place where you're errorring from 👍.
 
> Do you think this error forwarding issue may be the root cause of why the `err.stderr` from the `execute` test helper doesn't include the thrown error?

Hmm, not sure really, would have to check. It's been a little bit, and I ended up going with a different approach in the end.

bernardobridge (approved):
Seems this is working appropriately after https://github.com/artilleryio/artillery/pull/2102 has been merged.

It also addresses https://github.com/artilleryio/artillery/issues/1693 and is a nice quality of life improvement 👍.

Merging this in.

Issue #2125, opened 2023-09-09
Regexp group match not working


sancag:
Version info: 
```
Artillery: 2.0.0-36
Node.js:   v20.3.0
OS:        darwin
```

Running this command: 
```
capture:
          - regexp: '(cart#)([A-Z0-9]+)'
            group: 2
            as: 'capturedSegmentAfterHash'
```

I expected to see this happen:

With the following captured input

```
{"expiresAt":1694239938,"sk":"meta","pk":"cart#01H9W886XP9NXFPCXP57KQ4B48","isRefundableBookingSelected":false,"refundableAmount":0}}
```

I epect just the id `01H9W886XP9NXFPCXP57KQ4B48`

Instead, this happened:
I get `cart#01H9W886XP9NXFPCXP57KQ4B48` including `cart#`
```

capturedSegmentAfterHash: {
    value: 'cart#01H9W9A1N0H4SHJMSER4WZSRZ0',
    strict: undefined,
    failed: false
  }
```

sancag:
You are a ⭐️ @InesNi . Works perfectly

InesNi:
Glad I could help @sancag 🙌🏻

bernardobridge:
A fix for this is now released on [Artillery v2.0.0-38](https://github.com/artilleryio/artillery/releases/tag/v2.0.0-38).

PR #2127, opened 2023-09-12
fix(commons): match regex groups by integer index

InesNi:
# What

Fix for #2125 
The `extractRegExp` function was not considering the integer index of the group in the regular expression as a group identifier, only [named groups](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Named_capturing_group) were considered.

# How
Implemented the integer index consideration.

# Testing
Tested manually. As no tests exists for `engine_util` yet, the tests implementation will be added in a separate PR when created.

### DIFF
--- a/packages/commons/engine_util.js
+++ b/packages/commons/engine_util.js
@@ -609,12 +609,16 @@ function extractRegExp(doc, expr, opts) {
     return '';
   }
 
+  // Captures named group (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Named_capturing_group)
   if (group && match.groups) {
     return match.groups[group];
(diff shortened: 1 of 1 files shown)


packages/commons/engine_util.js
```diff
@@ -609,6 +611,8 @@ function extractRegExp(doc, expr, opts) {
 
   if (group && match.groups) {
```
bernardobridge, line 612: Hey @InesNi ! Looks good!

Could you please add some comments above the logic explaining the different forks, since we don't have tests for it yet?

bernardobridge (changes requested):

bernardobridge (approved):

PR #2135, opened 2023-09-13
fix(schema): correct the plugins allowed in root config

bernardobridge:
## Why

Not all built-in plugins allow you to use `config.<plugin-name>` to set config options. As far as I know, only `ensure` and `apdex` do. The schema assumed that all of them did.

### DIFF
--- a/packages/types/schema/plugins/publish-metrics.js
+++ b/packages/types/schema/plugins/publish-metrics.js
@@ -187,24 +187,26 @@ const OpenTelemetryReporterSchema = Joi.object({
   .unknown(false)
   .meta({ title: 'OpenTelemetry Reporter' });
 
-const PublishMetricsPluginConfigSchema = Joi.array().items(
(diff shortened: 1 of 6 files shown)


packages/types/schema/plugins/publish-metrics.js
```diff
@@ -187,24 +187,26 @@ const OpenTelemetryReporterSchema = Joi.object({
...
+        HoneycombReporterSchema,
+        LightstepReporterSchema,
+        MixpanelReporterSchema,
+        StatsdReporterSchema,
+        InfluxReporterSchema,
+        OpenTelemetryReporterSchema
+      )
+      .match('one')
+  )
+  .meta({ title: 'Publish Metrics Plugin' });
```
bernardobridge, line 209: this is just formatting - I added titles to all the plugins to help differentiate them visually in the configs.

PR #2139, opened 2023-09-13
Socket.io engine fixes + more flexible `response.data` behavior

josh-coderpad:
The commits give more details, but in summary this PR fixes a couple of bugs around response matching so the tool behaves as the schema/docs indicate it should.

I've also broadened what `response.data` will accept (https://github.com/artilleryio/artillery/commit/22f1e569654230b47a8cde15390697e69a571aea), because:
- This currently wasn't supported
- As this feature was broken, it's unlikely anyone's depending on the current strict behavior
- IMO it's a good compromise between simplicity (no change to config) and additional functionality (matching multiple responses + substring instead of exhaustive match). Alternatives would include allow regular expressions + a specific setting for multiple messages, but I think this is a good starting point 😄 

Please let me know if I can provide more info, and thanks for this excellent tool!

### DIFF
--- a/packages/artillery/lib/util/validate-script.js
+++ b/packages/artillery/lib/util/validate-script.js
@@ -53,6 +53,7 @@ const socketioItems = {
     then: Joi.alternatives(
       Joi.object({
         channel: Joi.string(),
+        concat: Joi.boolean(),
         data: Joi.any()
       }),
       Joi.array().items(Joi.string())
@@ -86,7 +87,7 @@ const flowItemSchema = Joi.object({
   }),
   otherwise: Joi.when('...engine', {
     is: Joi.exist().valid('socketio'),
-    then: Joi.object().max(3),
+    then: Joi.object().max(4),
     otherwise: Joi.object().length(1)
   })
 });
--- a/packages/core/lib/engine_socketio.js
+++ b/packages/core/lib/engine_socketio.js
@@ -61,19 +61,35 @@ function isAcknowledgeRequired(spec) {
   return spec.emit && spec.acknowledge;
 }
 
-function processResponse(ee, data, response, context, callback) {
-  function isValid(data, response) {
-    if (!Array.isArray(response.data)) {
-      //`json` key is added at some point to the response.data object, to use with `captureOrMatch` function
-      //we should omit it when comparing the response to the data
-      const responseDataWithoutJson = _.isObject(response.data)
-        ? _.omit(response.data, 'json')
-        : response.data;
-      return deepEqual(data[0], responseDataWithoutJson);
-    }
-
+function isValid(data, response) {
+  if (_.isArray(response.data)) {
+    //we check if it's an array first (as arrays are objects), and if it's an array, do a deep equality check between both arrays
     return deepEqual(data, response.data);
   }
+
+  if (_.isObject(response.data)) {
+    //`json` key is added at some point to the response.data object, to use with `captureOrMatch` function
+    //we should omit it when comparing the response to the data
+    const expectedResponse = _.omit(response.data, 'json');
+    const actualResponse = data[data.length - 1]; // if response.data is not an array, we compare it to the last element of the actual response
+
+    return deepEqual(actualResponse, expectedResponse);
+  }
+
+  if (_.isString(response.data)) {
+    const expectedResponse = response.data;
+    const actualResponse = data[data.length - 1]; // if response.data is not an array, we compare it to the last element of the actual response
+    debug(
+      `checking if string ${expectedResponse} is a partial match for string ${actualResponse}`
+    );
+    return actualResponse.includes(expectedResponse); //we accept a partial match if it's a string
+  }
+
+  debug(`unexpected data type for response.data: ${typeof response.data}`);
+  return false;
+}
+
+function processResponse(ee, data, response, context, callback) {
   // Do we have supplied data to validate?
   if (response.data && !isValid(data, response)) {
     debug('data is not valid:');
@@ -237,7 +253,7 @@ SocketIoEngine.prototype.step = function (requestSpec, ee) {
           socketio.emit(...outgoing);
         }
         markEndTime(ee, context, startedAt);
-        return callback(null, context);
+        return callback(err, context);
       }
     }; // endCallback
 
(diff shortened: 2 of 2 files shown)


bernardobridge:
Hi @josh-coderpad ! First off, thanks or contributing to Artillery, especially with such a well thought PR! 🙇‍♂️ 

A couple of things:
- I think the CLA is still pending to be signed, have you been able to sign it yet?
- Regarding https://github.com/artilleryio/artillery/pull/2139/commits/236710857ac045ccc19ee76a60a38cd6e62ede11 , i.e. changing `response` to be under `emit`. If you look over the `git blame` on that code, this was actually an intended change from this PR https://github.com/artilleryio/artillery/pull/1112. I believe this was to bring it in-line with a popular third-party Artillery engine. However, it seems that our documentation may be outdated, and perhaps that is why the feature appears to not be working for you? (i.e. `response` should not be under `emit`, but in its own key)
- As for the rest of the changes, I think we'd first need to revert the `response` change above (as otherwise the whole socketio test suite will fail I believe), as we'd want the test suite to run successfully first. 

I'll also try to discuss those changes internally with the team, but bear with us a little bit as there's some PTO going on at the moment :)

josh-coderpad:
Hey @bernardobridge!

Sorry for the radio silence. Busy end of quarter here.

Thanks for your comments, I indeed was following the socketio engine docs on your site to lead me to this fix, and definitely think `response` makes sense as a sibling to `emit`. I've rebased my MR and signed the CLA for this account/repo. Let me know if the changes work for you.

josh-coderpad:
One comment on the new code: it'd be nice to be able to specify:
```yaml
    flow:
      - response:
          channel: serviceStatus
          data: ready
```
rather than:
```yaml
    flow:
      - emit:
          channel: ignoreme
          data: ignore this (sent from load tests)
        response:
          channel: responseChannel
          data: some data I get automatically without needing to send anything
```

I looked through the code a bit and it seems it'd need a bit of cleanup first, as it assumes `emit` is always present in a few places/callbacks.

josh-coderpad:
Hi @bernardobridge, did you have a chance to review the changes? I'm happy with it in its current state. Let me know if you want me to open a ticket for [this](https://github.com/artilleryio/artillery/pull/2139#issuecomment-1741425517).

bernardobridge:
> Hi @bernardobridge, did you have a chance to review the changes? I'm happy with it in its current state. Let me know if you want me to open a ticket for [this](https://github.com/artilleryio/artillery/pull/2139#issuecomment-1741425517).

Hi @josh-coderpad ! I'm so sorry, I somehow missed your previous message (got buried in email notifications!).

I've discussed internally and the change makes sense to us, and thank you again for thinking things through. We'd like to get this in before next release, together with [this PR](https://github.com/artilleryio/artillery/pull/2171) that I don't know if you've had a chance to take a look. I was wondering if that also helps at all with your use case? 

The only thing is I'd like to make sure we add automated tests for this - we're trying very hard now to keep our coverage up. I should be able to look into adding some tests for it by the end of the week, so if you don't mind I can contribute directly here. Otherwise if you have time and want to take a crack at it, feel free to go for it too!

---

Regarding your other feature request, for now we won't have the bandwidth to work on that as we have a few things in other engines/features that are higher priority right now. So yes, please open at as a separate issue and we'll track it as a future enhancement (and we can discuss there the use case). 👍

josh-coderpad:
I checked out that MR and noticed it was merged recently. Rebased onto main and fixed things locally.

I've opened [an issue for the scenario syntax above](https://github.com/artilleryio/artillery/issues/2286), as well as [one for `DEBUG=socketio`](https://github.com/artilleryio/artillery/issues/2287).

Tests sound great! I likely won't have time soon, but will circle back if I do. Feel free to use this branch/MR. I'd also like to get this in the next release to avoid hosting it internally... so let me know if you need more info from me.

bernardobridge:
Hi @josh-coderpad 👋 ! Thanks again for this. 

The current test suite is failing in a socket-io test. I've had a chance to look at it, and I believe the reason is that the current change assumes `responseData` is a string, but that's not necessarily the case (it could be an array, or an object). I think we should still be able to get your feature in. The main conflict will be with the whole `Matching a substring of the response, rather than the entire response` part. I think the reasonable thing to do is to check if it's a string, and only apply this logic in that case.

I've taken a stab at it, and the final change would look something like this:  https://github.com/artilleryio/artillery/compare/socketio-new-changes. The main change is:
- Push all args to `responseData`, rather than a concatenation of them;
- If it's specifically a string type, then we still allow a substring match. Otherwise, it maintains the previous functionality (deep equality checks for an object and array);

This also passes the existing test suite (plus I tested a simulated example of receiving a partial response x seconds later, and it worked). If you're happy to incorporate those changes, then I can help get it over the line with a test.

Let me know!

josh-coderpad:
@bernardobridge hey thanks for taking a stab at it!

I like you changes and agree with the approach. Unfortunately for us we have a bit of a strange case where we're sending data back as we get it, so the responses end up a little garbled sometimes (e.g. `['som', 'e dat', 'a\n']`).

I understand depending on the use case it doesn't necessarily make sense to consider the entire response a string, so maybe a good compromise would be to make that opt in? Either with a specific boolean option (something like `concat`), or with a "bring your own validation function".

I've pushed your changes along with a commit that adds `response.concat`, which seems to pass the tests, and works for me use case as well. What do you think about this option as a middle ground?

josh-coderpad:
I think the remaining test is missing `$CLOUD_KEY` ([here](https://github.com/artilleryio/artillery/blob/18ad4acd9201eb52bc236bcebee1a9146bee9c01/.github/workflows/examples.yml#L33)), but I don't think I have access to set that. Let me know if this is something I can fix though.

bernardobridge:
Hi @josh-coderpad,

That's a current limitation of those example runs - they can't run against forks. We'll fix that soon.

I spoke with the team, and let's go ahead with the change as is. I think there is only one small thing you may have to fix - since you added a field at the same level as `channel`, I believe this needs to be updated to max(4): https://github.com/artilleryio/artillery/blob/18ad4acd9201eb52bc236bcebee1a9146bee9c01/packages/artillery/lib/util/validate-script.js#L89 . Otherwise if `namespace` is set too, it will fail that validation. 

I know `namespace` doesn't appear on the schema there, but that schema is slightly out of date, and will be replaced eventually with a much more comprehensive one we have.

So if you just fix that validation, we can go ahead and merge, and I'll add tests for this at a later time. This way it can be released on the next version which we're looking to release this week.

josh-coderpad:
Got it, updated to allow 4 items there. Thanks for all the back and forth on this one!

bernardobridge:
Hey @josh-coderpad. Thanks so much as well for the contribution and sticking with the back and forth 👍 . I think we're ready to merge this now! Going to go ahead and do it now.

I've also fixed the examples against forks problem for now, so should be fine on future contributions 🎉 .

As I said, I'll make test(s) for this as soon as I have the time, and will also update our docs.

Issue #2141, opened 2023-09-15
Update readme URL's to point to non 404 URL's


dirtydiesel46:
Version info: 
```
v2.0.0-37
```
<command>

I expected to see this happen:

When routing to https://github.com/artilleryio/artillery/tree/main/packages/artillery#docs-and-guides and clicking on [Running tests using real headless browsers](https://docs.art/playwright) - I was expecting it to [Why load test with headless browsers?
](https://www.artillery.io/docs/reference/engines/playwright#why-load-test-with-headless-browsers)

![image](https://github.com/artilleryio/artillery/assets/48984939/0293841c-f7f0-4a14-abc2-a92840b750ec)

**Update** 

All of the URL's in below image are outdated ( and renaming this Issue) - hence the PR should fix all:

![image](https://github.com/artilleryio/artillery/assets/48984939/b3dcffa4-b819-47d2-9815-2f8b817c4166)

Instead, this happened:

Instead it routed to a 404 error page - 

```
404 | This page could not be found.
```

![image](https://github.com/artilleryio/artillery/assets/48984939/f53a2c49-0a0b-4be3-95ee-688b956a3649)

Files being used:

[readme.md](https://github.com/artilleryio/artillery/blob/main/packages/artillery/README.md)

InesNi:
Hi @dirtydiesel46 ,

Thank you for reporting this and submitting a PR to fix it. The url's were correct, but the redirects were broken. It's all fixed now 👍🏻, so we closed the PR.

Thanks again 🙏🏻 !

PR #2143, opened 2023-09-15
fix: Update URL's to artillery.io/docs

dirtydiesel46:
This fixes https://github.com/artilleryio/artillery/issues/2141
fix: Update URL's to artillery.io/docs

Looked at all the Learn more section Docs and guides and fixed the URL's to point to working URL's rather than 404's


kemaldotearth:
Hi @dirtydiesel46 this should now be fixed with the shortened URL's. Thanks for reporting! :)

PR #2153, opened 2023-09-19
feat(publish-metrics): implement otlp grpc trace exporter

InesNi:
# What

Base implementation of OTLP/gRPC trace exporter for OpenTelemetry reporter. Authentication, if needed, can be set through `headers` or environment variables, same as with other OpenTelemetry reporter exporters.
`headers`, if set in test script, will be passed to exporter as `metadata` key, same  as in the case of OTLP/gRPC metric exporter.

Environment variable for exporter headers is `OTEL_EXPORTER_OTLP_METRICS_HEADERS`.


bernardobridge (approved):
:shipit:

Issue #932, opened 2021-02-19
The context in beforeScenario/afterScenario doesn't contain the scenario name


Shazwazza:
It would be helpful if the context or event data for beforeScenario/afterScenario contained the scenario name. I am using these methods to initialize and clear custom code for streaming event counters and need to record for which scenario it is.

hassy:
good shout @Shazwazza, those hooks should probably have access to the entire scenario object

bernardobridge:
This feature is now released in [Artillery v2.0.0-38](https://github.com/artilleryio/artillery/releases/tag/v2.0.0-38).

You have access to the whole scenario object through `context.scenario`. For example:

```
function myBeforeScenarioHandler(context, ee, next) {
    //access context.scenario to obtain any property of the scenario script
    ee.emit('counter', `my_custom_counter_${context.scenario.name}`);
    next();
};
```

PR #2158, opened 2023-09-21
feat(core): include scenario object in engine hooks

bernardobridge:
## Why

Addresses https://github.com/artilleryio/artillery/issues/932

It's useful for test authors to have access to the current scenario being ran for multiple reasons, one of them being the use case explained in that issue.

---

_Note: removed unrelated `index.js` from tests that was no longer being used after test change in another PR._


packages/core/lib/engine_http.js
```diff
@@ -167,7 +167,7 @@ HttpEngine.prototype.createScenario = function (scenarioSpec, ee) {
     });
   });
 
-  return self.compile(tasks, scenarioSpec.flow, ee);
+  return self.compile(tasks, scenarioSpec, ee);
```
bernardobridge, line 170: `compile` was being passed `scenarioSpec.flow` but not doing anything with it, therefore the change here doesn't really change anything, and we can now use the whole object.

PR #2163, opened 2023-09-22
feat(core): add mean to calculated histogram metrics

bernardobridge:
## Why

Some users may be interested in observing the difference between `mean` and other statistics, as we've received as feedback. As we expose a lot of other statistics, we can easily expose this one too.

---

_Open question: should we change `ensure` to include this in the legacy conditions?_


bernardobridge:
@hassy , it's available in the report now: 

Screenshot 2023-10-10 at 14 18 16

PR #2171, opened 2023-09-26
feat(socketio): add args+on for event validation of any amount of args

bernardobridge:
## Description

A porting of the popular (and really good) third-party engine (https://github.com/ptejada/artillery-engine-socketio-v3) to Artillery Core was done back in July 2021 in https://github.com/artilleryio/artillery/pull/1112.

However, shortly after that, updates were done to the engine that were never ported over to Artillery, which left the job incomplete. This update in particular makes a lot of sense, because it gives support for multiple arguments to be easily asserted on, which is something that can be done when emitting SocketIO events.

Essentially, it gives the ability to do things like:

```yaml
      - emit: ["message", "chat_room_1", "Hey there!"]
        response:
          on: message
          args:
            - chat_room_1
            - Hey there!
```

## Pre-merge checklist

- [] Does this require an update to the docs? _**Yes**_
- [] Does this require a changelog entry? _**Yes**_

### DIFF
--- a/packages/core/lib/engine_socketio.js
+++ b/packages/core/lib/engine_socketio.js
@@ -52,7 +52,9 @@ function markEndTime(ee, _, startedAt) {
 }
 
 function isResponseRequired(spec) {
-  return spec.emit && spec.response && spec.response.channel;
+  return (
+    spec.emit && spec.response && (spec.response.channel || spec.response.on)
+  );
 }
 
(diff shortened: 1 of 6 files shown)


bernardobridge:
Merging this as a merge commit rather than squash, because the last commit introduces a deviation (fix) from the third-party engine, and for ease of search, I'd like that reflected in the commit history.

Issue #2147, opened 2023-09-18
Issues with multiple concurrent users


mjohnson767:
Hello Hassy, I'm very interested in the Artillery | Playwright testing but experiencing the same issue as # 1711.
Build: artillery/2.0.0-37 win32-x64 node-v18.14.0
My goal is run tests with the VUser concurrent limit with a ramp-up (to the limit) and duration for the test. This is pretty much standard for all load testing tools. For some quick testing I've set a low amount of Vusers to test it out. With the following setup the actual concurrent rate is only 1 VUser that just repeats until the duration is hit.

config:
target: "https://mytesturl.com/"
phases:
- duration: 60
rampTo: 3
name: Ramp up load
maxVusers: 3
payload:
- path: ./logins.csv
fields:
- "userid"
- "recordid"
order: sequence

If I take out the maxUsers I'll see several browsers open with only one of the actually going through the workflow.

Attempting to run with the WORKERS setting fails with this version. WORKERS=3 artillery run script.yaml
WORKERS=3 : The term 'WORKERS=3' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included,
verify that the path is correct and try again.

Do you have any ideas?
I've ran tests with Locust/Selenium locally and I know I can 20-30 browsers on my laptop. Running Artillery/Playwright should easily out perform locust/selenium. With a single load generator we were able to run 175 headless browsers. I'm hoping to surpass that with Artillery/Playwright.

### Discussed in https://github.com/artilleryio/artillery/discussions/1711

<div type='discussions-op-text'>

<sup>Originally posted by **JulieMatic** December  5, 2022</sup>
I am currently working on some tests that are using artillery-engine-playwright, and I'm running into a couple of different issues depending on the combination of versions I use. 

### Only the "active" Chrome window is interacted with
```
phases:
            - name: 10user
              duration: 300
              arrivalRate: 1
              arrivalCount: 1000
              maxVusers: 10
```
[...]

PR #2173, opened 2023-09-27
feat(playwright): add step transaction function to user context for custom metrics

bernardobridge:
## What

Gives users an abstraction to facilitate breaking their Playwright scenario into a series of steps and capture transaction timers for each of them, with an API that should be familiar to Playwright users.

### Example (based on real user scenario)
```js
async function loginSearchAndLogout(page, userContext, events, test) {
   //1. simply add this line to your scenario function, or use test.step below instead
  const { step } = test;
  const userid = userContext.vars.userid;
  const recordid = userContext.vars.recordid;

  //2. wrap any logic you have in steps 
  //(sometimes you might already have something like this done from existing playwright tests)
  await step("landing_page", async () => {
    await page.goto('https://internaltesturl.com/landing');
    await page.getByLabel('id-label').fill( `${userid}` );
    await page.getByLabel('Password').fill(`${password}`);
  })

  await step("submit_login", async () => {
    await page.getByRole('button', { name: 'Submit' }).click();
    await page.getByPlaceholder('enter request id').fill(`${recordid}`);
  })

  await step("search_record", async () => {
    await page.getByRole('button', { name: 'Go' }).click();
    await page.locator('css=button.avatar-button').click();
  })

  await step("logout", async () => {
    await page.getByText('Logout').click();
  })
}
```

This will leverage Artillery's custom metrics to create metrics that look like this:
```
browser.step.landing_page:
  min: ......................................................................... 87
  max: ......................................................................... 150
  median: ...................................................................... 89.1
  p95: ......................................................................... 89.1
  p99: ......................................................................... 89.1
browser.step.submit_login:
  min: ......................................................................... 300
  max: ......................................................................... 716
  median: ...................................................................... 561.2
  p95: ......................................................................... 561.2
  p99: ......................................................................... 561.2
browser.step.search_record:
  min: ......................................................................... 287
  max: ......................................................................... 801
  median: ...................................................................... 290.1
  p95: ......................................................................... 290.1
  p99: ......................................................................... 290.1
browser.step.logout:
  min: ......................................................................... 52
  max: ......................................................................... 334
  median: ...................................................................... 140.2
  p95: ......................................................................... 200.4
  p99: ......................................................................... 200.4
```

- Web vital metrics give you a single pane of glass look into performance of individual pages.
- Steps give you control to break it down into the transactions you care about and analyse their performance over time.

## Why

This pattern was already present in the examples, and recently there was demand for it. After abstracting it into a utility function (https://github.com/artilleryio/artillery/issues/2147#issuecomment-1729259684 and https://github.com/artilleryio/artillery/discussions/2164#discussioncomment-7102210), and the received feedback, it became obvious that users could benefit from this.

## Testing

Tested against the examples (which will be updated once this is released into the next version).

### DIFF
--- a/packages/artillery-engine-playwright/index.js
+++ b/packages/artillery-engine-playwright/index.js
@@ -40,6 +40,14 @@ class PlaywrightEngine {
       return self.aggregateByName && spec.name ? spec.name : url;
     }
 
+    const step = async (stepName, userActions) => {
+      const startedTime = Date.now();
+      await userActions();
+      const difference = Date.now() - startedTime;
+
+      events.emit('histogram', `browser.step.${stepName}`, difference);
+    };
+
     return async function scenario(initialContext, cb) {
       events.emit('started');
       const launchOptions = Object.assign(
@@ -74,7 +82,7 @@ class PlaywrightEngine {
                   name: metric.name,
                   value: metric.value,
                   metric: metric,
-                  url: window.location.href
+                  url: window.location.href // eslint-disable-line
                 })
               );
             });
@@ -91,8 +99,8 @@ class PlaywrightEngine {
           }
 
           try {
-            const performanceTimingJson = await page.evaluate(() =>
-              JSON.stringify(window.performance.timing)
+            const performanceTimingJson = await page.evaluate(
+              () => JSON.stringify(window.performance.timing) // eslint-disable-line
             );
             const performanceTiming = JSON.parse(performanceTimingJson);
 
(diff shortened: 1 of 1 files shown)


packages/artillery-engine-playwright/index.js
```diff
@@ -74,7 +82,7 @@ class PlaywrightEngine {
                   name: metric.name,
                   value: metric.value,
                   metric: metric,
-                  url: window.location.href
+                  url: window.location.href // eslint-disable-line
```
hassy, line 85: curious - what does eslint complain about here? (and elsewhere)

hassy:
lgtm

packages/artillery-engine-playwright/index.js
bernardobridge: It complains about this:
```
   85:24  error    'window' is not defined               no-undef
```

I'll disable the specific rule actually on all of these, to make it more clear 👍

bernardobridge: It still complains about `webVitals` btw:
`   80:13  error    'webVitals' is not defined            no-undef`

But for that one there's no way for eslint to disable it due to being inline, IIRC. We'll have to move the webVitals script to its own file at some point in another PR to address this

PR #2174, opened 2023-09-28
feat(playwright): add trace support with opentelemetry

InesNi:
## What

Basic tracing support for Playwright Engine with OpenTelemetry, in a form of a function that is made available to the user through the `test` - the last argument of the user flow function.
This function is the tracing equivalent to the [`step` function](https://github.com/artilleryio/artillery/pull/2173), and is used instead of it in the case where OpenTelemetry reporter is configured for tracing (within the `publish-metrics` plugin).

If some of the test scenarios are configured for the `HTTP` engine, they will be traced as normal along with the Playwright scenarios.

### From engine side 
- Playwright engine checks if Otel tracing is configured, and if it is it calls the `beforeScenario` hook function with the following arguments `page`, `initialContext`, `events` and the User's `processor function`.

### From reporters side: 
- When tracing is configured, reporter looks for engines used in the script, and if `Playwright` engine is found it creates a `beforeScenario`, `afterScenario` and  the `traceStepFunction` hook.
- Scenario span is started right before we call the user `flow` function with the `beforeScenario` hook
- Check is made to see if tracing is configured, and if it is, we pass this tracing `step` function instead of the regular one to the user. 
- The scenario span is provided as a parent to the step function, so that all the spans user creates are going to be nested inside it.
- Once the user function is done we end the scenario span.

- The `step` function is responsible for starting and ending a child span with user configured name, optional `attributes` and whatever else the user sets on the span within their processor.
-  It also emits a custom `browser.step.{stepName}` histogram so it can use interchangeably with the non tracing `step` function as proposed [in the comments](https://github.com/artilleryio/artillery/pull/2174#discussion_r1340182493)

## Configuration
**testScript.yaml**

```yaml
config:
  target: "https://www.artillery.io"
  phases:
    - arrivalRate: 5
      duration: 12
  engines:
    playwright: {}
  plugins:
    publish-metrics:
      - type: 'open-telemetry'
        serviceName: "Artillery-Playwright-test"
        traces:
          exporter: otlp-proto
          endpoint: "{{ $processEnvironment.OTEL_EXPORTER_OTLP_TRACESS_ENDPOINT }}"
  processor: ./myProcessor.js
scenarios:
  - name: "Some scenario"
    engine: playwright
    flowFunction: "someUserFunction"

```

**myProcessor.js**

```javascript
async function someUserFunction(page, vuContext, events, test) {

  await test.step('Going to website', async (span) =>{
    const response = await page.goto('https://www.artillery.io/');
    span.setAttributes({
      'status_code', response.status(),
      'url.full': response.url())
    const request = await response.request()
    const timings = await request.timing()
    span.setAttributes(timings)
  }, {'MyCustomAttribute': 'Custom value', step: 1})  // Custom attributes as the last optional argument

  await test.step('Navigate To Cloud', async (span) =>{
    await page
    .getByLabel('Main navigation')
    .getByRole('link', { name: 'Cloud' })
    .click();
    span.addEvent('click')
  }, {step: 2})

  await test.step('Click button to join', async (span) =>{
    await page
    .getByRole('button', { name: 'Join Artillery Cloud early access waitlist' })
    .click();
  })
```

## Results
The following shows a trace in Honeycomb:
Screenshot 2023-09-28 at 12 28 37

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
+++ b/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
@@ -27,7 +27,7 @@ class OTelReporter {
       process.env.DEBUG &&
       process.env.DEBUG === 'plugin:publish-metrics:open-telemetry'
     ) {
-      diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.ERROR);
+      diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
     }
     this.metricExporters = {
       'otlp-proto'(options) {
--- a/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
+++ b/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
@@ -11,6 +11,7 @@ const {
   SpanKind,
   SpanStatusCode,
   trace,
+  context,
   metrics
 } = require('@opentelemetry/api');
 const { Resource } = require('@opentelemetry/resources');
@@ -26,7 +27,7 @@ class OTelReporter {
       process.env.DEBUG &&
       process.env.DEBUG === 'plugin:publish-metrics:open-telemetry'
     ) {
-      diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
+      diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.ERROR);
     }
     this.metricExporters = {
       'otlp-proto'(options) {
@@ -115,6 +116,7 @@ class OTelReporter {
     }
 
     if (config.traces) {
+      // Set basics needed regardless of the engine 
       this.traceConfig = config.traces;
       this.validateExporter(
         this.traceExporters,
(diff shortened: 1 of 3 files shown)


packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -395,6 +416,62 @@ class OTelReporter {
...
+      if (configAttributes) {
+        span.setAttributes(configAttributes);
+      }
+      await callback(span);
+      if (attributes) {
+        span.setAttributes(attributes);
+      }
+      span.end();
+    };
+  }
```
bernardobridge, line 473: Hey @InesNi 👋 ! Nice work on this, it's great that this is API is compatible with the one introduced in the PR here https://github.com/artilleryio/artillery/pull/2173/files - it will make it easy for the user.

However, in the event that the user has already wrapped their test with steps using that method, they would now need to change their test to use `traceStep`. Ideally, they would just need to enable the publish-metrics plugin. 

I'm wondering if it might not be better to call this function `step` as well, and have it override my `step` function when tracing is enabled. And then you can have your function emit custom timers too. This would make the transition for the user transparent - they would just have the optional configuration available for them of setting `attributes` if desired.

bernardobridge (changes requested):

packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
InesNi: Oh I see, good thinking! 🙌🏻 I will change the name and have the function emit the timers then for now. I would also turn this PR into a draft so I can adjust the behaviour once your PR has been merged, if you think that makes sense @bernardobridge?

PR #2184, opened 2023-10-03
feat(artillery): add ability to specify scenario to run by name

bernardobridge:
## Why

Originally requested here https://github.com/artilleryio/artillery/discussions/1929, this is a useful feature for people who are saving multiple scenarios in the same file. Especially useful once you start having complex scenarios with weights, but still want the ability to run them individually on demand, a requirement I've had myself in the past.

~Note: CLI flag name (`--name`) could be changed to something else, like `--byName`. Please suggest something else if you think it makes more sense.~
_Note: Flag name is now --scenario-name after some internal discussions. Please make other suggestions if other ideas pop up._

### DIFF
--- a/packages/artillery/test/cli/command-run.test.js
+++ b/packages/artillery/test/cli/command-run.test.js
@@ -108,7 +108,10 @@ tap.test('Can specify scenario to run by name', async (t) => {
 
   t.ok(
     deleteFile(reportFilePath) &&
-      json.aggregate.counters['vusers.created_by_name.Test Scenario 2'] === 6
+      json.aggregate.counters['vusers.created_by_name.Test Scenario 2'] === 6 &&
+      typeof json.aggregate.counters[
+        'vusers.created_by_name.Test Scenario 1'
+      ] === 'undefined'
   );
 });
 
--- a/packages/artillery/test/cli/command-run.test.js
+++ b/packages/artillery/test/cli/command-run.test.js
@@ -88,6 +88,32 @@ tap.test('Environment specified with -e should be used', async (t) => {
   t.ok(exitCode === 0 && !output.stdout.includes('ECONNREFUSED'));
 });
 
+tap.test('Can specify scenario to run by name', async (t) => {
+  const reportFile = 'report-with-scenario-by-name.json';
+  const reportFilePath = await getRootPath(reportFile);
+
+  const [exitCode, output] = await execute([
+    'run',
+    '-n',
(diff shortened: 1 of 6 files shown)


bernardobridge:
@hassy made the changes we discussed. I went with erroring from within here https://github.com/artilleryio/artillery/pull/2184/commits/a5604aeca792bf1be9bdc62a4dc656a07c472a1b (i.e. worker level), rather than earlier in the script. The reason is that this way it's not dependant on `run` command implementation, which means it should work with Fargate (which doesn't have access to the test script from within the command). This is also where the logic of picking is, so it felt natural to have that there for now. When Fargate moves over, we can revisit that IMO.

Screenshot of error:
Screenshot 2023-10-06 at 12 44 43

packages/artillery/test/cli/command-run.test.js
```diff
@@ -88,6 +88,49 @@ tap.test('Environment specified with -e should be used', async (t) => {
...
+    'test/scripts/scenario-named/scenario.yml'
+  ]);
+
+  t.ok(
+    exitCode === 0 && output.stdout.includes('Successfully running scenario 2')
+  );
+  const json = JSON.parse(fs.readFileSync(reportFilePath, 'utf8'));
+
+  t.ok(
+    deleteFile(reportFilePath) &&
```
hassy, line 110: Not for this PR but we should not need to do this. The file should just be created in a temporary location (under `/tmp`) instead.

packages/artillery/test/cli/command-run.test.js
```diff
@@ -88,6 +88,49 @@ tap.test('Environment specified with -e should be used', async (t) => {
...
+      json.aggregate.counters['vusers.created_by_name.Test Scenario 2'] === 6
```
hassy, line 111: This works to make sure that no other scenarios were run but would be nice to have a more explicit test for it.

```suggestion
      typeof json.aggregate.counters['vusers.created_by_name.Test Scenario 1'] === 'undefined'
```

packages/artillery/test/cli/command-run.test.js
bernardobridge: I agree 👍, I was implementing the same pattern that was done throughout, but I can make a separate PR getting rid of this pattern

bernardobridge:
I've made this work with Lambda now, and tested it 👍. Both happy path and errors work now.

Fargate support will require updating platform-fargate, which I'll do next.

PR #2188, opened 2023-10-05
feat(playwright): trace user flow through performance api

InesNi:
# What

Tracing support for Playwright Engine through OpenTelemetry reporter (`publish-metrics` plugin). 

This tracing support is implemented not through `request`-`response` events/object but through the use of [Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance_API).

Traces are created for each `scenario`. The scenario is a parent span, within it there is a child span for each [`performanceEntry`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry) of [`resource`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) and [`navigation`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming) type. Then each entry has child spans that represent each of the important timing events:

Screenshot 2023-10-05 at 17 03 31

## Limitations
There are certain limitations to take in consideration: 
- Traces do not have the 'request.method' information
- `spanStatus` set as `ERROR` only if `response status code` is 400 or more
- An additional limitation seems to be that at times certain requests might be missed due to the speed of execution.

No additional/special configuration is needed, when OTel reporter is in use for tracing with the Playwright engine scenarios configured, it will automatically trace any Playwright, as well as HTTP engine scenarios.
# How

A scenario hook `traceFlowFunction` is used to hook the OTel reporter to the Playwright engine. 

The `performance.getEntries()` on the last `framenavigated` event gets the array of entries needed to set the spans which is stored in the `entries` variable outside of the eventListener callback. This way it is available to pass into the OTel reporters `tracePerformanceFlow` function that is called after the user flow function has finished and we have all the entries available. There we set all the spans manually with data available in the `entries`.

### DIFF
--- a/packages/artillery-engine-playwright/index.js
+++ b/packages/artillery-engine-playwright/index.js
@@ -207,10 +207,12 @@ class PlaywrightEngine {
         await fn(page, initialContext, events);
 
         // After the user flow function has finished we will have all the performance entries needed to set the spans
-        const tracePerformanceEntries = self.processor[spec.traceFlowFunction]
-        tracePerformanceEntries(entries, spec.name)
-        await page.close();
+        if (self.tracing){
+          const tracePerformanceEntries = self.processor[spec.traceFlowFunction]
+          tracePerformanceEntries(entries, spec.name)
+        }
 
+        await page.close();
 
         if (cb) {
           cb(null, initialContext);
--- a/packages/artillery-engine-playwright/index.js
+++ b/packages/artillery-engine-playwright/index.js
@@ -5,6 +5,8 @@ class PlaywrightEngine {
   constructor(script) {
     debug('constructor');
     this.target = script.config.target;
+    this.tracing = (script.config?.plugins?.['publish-metrics'] || []).some((config) => {
+      return (config.type === 'open-telemetry') && config.traces})
 
     this.config = script.config?.engines?.playwright || {};
     this.processor = script.config.processor || {};
@@ -84,6 +86,21 @@ class PlaywrightEngine {
         const page = await context.newPage();
 
         debug('page created');
+        // For tracing we need the performance entry array from the last frame navigated (when all the performanceEntry objects have been added)
+        // The 'entries' variable is set outside the emitter and updated with each 'framenavigated' event
+        let entries
+        page.on('framenavigated', async(page)=> {
+          if(self.tracing){
+            try{
+              const perfEntriesJSON = await page.evaluate(() => JSON.stringify(window.performance.getEntries()))
+              entries = JSON.parse(
+                perfEntriesJSON
+              )
+            }catch(err){
+              throw new Error(err)
+            }
+          }
+        })
 
         page.on('domcontentloaded', async (page) => {
           if (!self.extendedMetrics) {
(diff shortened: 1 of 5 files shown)


packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -395,6 +424,117 @@ class OTelReporter {
...
+      DNS_lookup: { start: 'domainLookupStart', end: 'domainLookupEnd' },
+      TCP_handshake: { start: 'connectStart', end: 'connectEnd' },
+      TLS_negotiation: { start: 'secureConnectionStart', end: 'requestStart' },
+      Request: { start: 'requestStart', end: 'responseStart' },
+      Response: { start: 'responseStart', end: 'responseEnd' },
+      'DOM.content_loaded.event': {
+        start: 'domContentLoadedEventStart',
+        end: 'domContentLoadedEventEnd'
+      },
+      'load.event': { start: 'loadEventStart', end: 'loadEventEnd' }
```
bernardobridge, line 442: We should have a more consistent naming convention here. I can see in the Honeycomb screenshot it looks a bit inconsistent. Perhaps:

- `redirect`
- `fetch`
- `dns_lookup`
- `tcp_handshake`
- `tls_negotiation`
- `request`
- `response`
- `dom_content_loaded` (do we need event?)
- `load` (do we need event?) 

Using `snake_case` should be better as anyway a lot of them get converted to that standard it seems (it seems Honeycomb/OTeL did it behind the scenes - probably due to Prometheus conventions).

bernardobridge (changes requested):
Nice work on this! Excited to see this through - this isn't a comprehensive review yet, but noticed this to start with

packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -395,6 +424,117 @@ class OTelReporter {
...
+
+    // Start Parent Span
+    this.playwrightTracer.startActiveSpan(
+      specName || 'Scenario execution',
+      { kind: SpanKind.CLIENT, startTime: parentStartTime },
+      (parent) => {
+        performanceEntries.forEach((entry) => {
+          if (
+            entry.entryType != 'resource' &&
+            entry.entryType != 'navigation'
```
bernardobridge, line 464: I think it would be good to document the ones we are filtering and why. 

On that note, I believe that sometimes there may be a few additional [PerformanceEntry types](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/entryType) we may be interested in here (when available), namely: `largest-contentful-paint`, `first-input` and maybe `layout-shift`.

packages/artillery-engine-playwright/index.js
```diff
@@ -186,10 +203,15 @@ class PlaywrightEngine {
         const fn =
           self.processor[spec.testFunction] ||
           self.processor[spec.flowFunction];
+
         await fn(page, initialContext, events);
 
+        // After the user flow function has finished we will have all the performance entries needed to set the spans
+        const tracePerformanceEntries = self.processor[spec.traceFlowFunction]
+        tracePerformanceEntries(entries, spec.name)
```
bernardobridge, line 211: shouldn't this be called only when `self.tracing` is available too?

bernardobridge (changes requested):

packages/artillery-engine-playwright/index.js
InesNi: Oh, missed that! good catch🙌🏻

Issue #2193, opened 2023-10-06
Apdex can't read processor properly


lfox91:
Version info: 
```
Artillery: 2.0.0-37
Node.js:   v18.17.1
OS:        darwin

-> npm -g ls
├── artillery-plugin-apdex@1.0.0
├── artillery-plugin-ensure@1.2.0
├── artillery-plugin-metrics-by-endpoint@1.2.0
├── artillery@2.0.0-37
├── corepack@0.15.1
├── npm@9.6.7
├── ts-node@10.9.1
└── typescript@5.2.2

```

Running this command: 
```
artillery run activate-login/activate-login.yml
```

I expected to see this happen:

The test run

Instead, this happened:

```
TypeError: Cannot create property 'apdexAfterResponse' on string './activate-login.js'

=======
class ApdexPlugin {
  constructor(script, _events) {
    this.script = script;

    const t = script.config.apdex?.threshold || script.config.plugins.apdex?.threshold || 500;
    console.log("!!!!\n",script.config.processor,"\n!!!")  // Prints: ./activate-login.js
    if (!script.config.processor) {
      script.config.processor = {};
    }

    script.scenarios.forEach(function (scenario) {
      scenario.afterResponse = [].concat(scenario.afterResponse || []);
      scenario.afterResponse.push('apdexAfterResponse');
    });

    function apdexAfterResponse(req, res, userContext, events, done) {
      const total = res.timings.phases.total;
      if (total <= t) {
        events.emit('counter', METRICS.satisfied, 1);
      } else if (total <= 4 * t) {
        events.emit('counter', METRICS.tolerated, 1);
      } else {
        events.emit('counter', METRICS.frustrated, 1);
      }

      return done();
    }

    // processor is being interpreted as a string which can't have a property
    script.config.processor.apdexAfterResponse = apdexAfterResponse;
                              ^^^^^^^^

```
The ApdexPlugin seems unable to treat the processor as an object as it's expecting. Everything works if I comment out Apdex in the config.

Files being used:
```
//activate-login.yml
config:
  target: [redacted]
  processor: ./activate-login.js
  tls:
    rejectUnauthorized: true
  http:
    timeout: 10
    maxSockets: 5
  phases:
    - duration: 5
      arrivalRate: 1
      rampTo: 2
      name: Warm up phase
    # - duration: 60
    #   arrivalRate: 5
    #   rampTo: 10
    #   name: Ramp up load
    # - duration: 30
    #   arrivalRate: 10
[...]

PR #2195, opened 2023-10-09
fix(apdex): run apdex logic correctly in workers

bernardobridge:
## Why

Solves https://github.com/artilleryio/artillery/issues/2193. Apdex has actually had this issue all along when using a processor. 

It happens because in the parent thread, the processors are not available as an array, that's an implementation detail of workers only.

In the case of this specific plugin, part of the logic has to run only on the parent thread, and the rest of the logic has to run in workers.

## Testing

Automated testing was implemented.

### Before
The implemented test replicates the issue:
https://github.com/artilleryio/artillery/actions/runs/6455923510/job/17524361652?pr=2195

Screenshot 2023-10-09 at 12 24 29

### After
https://github.com/artilleryio/artillery/actions/runs/6456170500/job/17525111146?pr=2195
Screenshot 2023-10-09 at 13 21 16

### DIFF
--- a/packages/artillery-plugin-apdex/index.js
+++ b/packages/artillery-plugin-apdex/index.js
@@ -15,6 +15,53 @@ class ApdexPlugin {
   constructor(script, _events) {
     this.script = script;
 
+    if (
+      global.artillery &&
+      Number(global.artillery.version.slice(0, 1)) > 1 &&
+      typeof process.env.LOCAL_WORKER_ID === 'undefined'
+    ) {
+      debug('Running in a worker, registering apdex beforeExit hook');
+      global.artillery.ext({
+        ext: 'beforeExit',
+        method: async (testInfo) => {
+          if (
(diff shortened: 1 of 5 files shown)


packages/artillery-plugin-apdex/index.js
```diff
@@ -15,6 +15,53 @@ class ApdexPlugin {
   constructor(script, _events) {
     this.script = script;
 
+    if (
```
hassy, line 18: Better to return early here if conditions are met. That makes the diff much smaller too as most of the existing code stays in place rather than getting moved around.

bernardobridge: Unless I'm missing something, I can't do an early return, as something always needs to run inside the if. Half the logic needs to run in workers, the other half in the main thread.

But I reversed the if condition anyway, as I do think checking for workers instead makes it easier to read, is more in-line with other plugins and makes for a smaller diff like you said 👍 .

PR #2210, opened 2023-10-12
fix(apdex): show all metrics in report

InesNi:
Related to #2209

# What
Making `apdex` plugin show all 3 metrics in the report even if their count is 0 (even if no metrics were frustrated/satisfied/tolerated) just like we do for `vusers.failed`. This is needed so that when the `threshold` for the given `apdex` metric is set in the `ensure` plugin, the `ensure` does not fail the check for metrics whose count is 0.

### DIFF
--- a/packages/artillery-plugin-apdex/index.js
+++ b/packages/artillery-plugin-apdex/index.js
@@ -36,11 +36,7 @@ class ApdexPlugin {
 
       function apdexAfterResponse(req, res, userContext, events, done) {
         const total = res.timings.phases.total;
-        const counts = {
-          satisfied: 0,
-          tolerated: 0,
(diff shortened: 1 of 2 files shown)


packages/artillery-plugin-apdex/index.js
```diff
@@ -36,13 +36,21 @@ class ApdexPlugin {
 
       function apdexAfterResponse(req, res, userContext, events, done) {
         const total = res.timings.phases.total;
+        const counts = {
```
hassy, line 39: Personal preference, but I'd emit all 3 metrics with value of `0` here at the beginning of the function, and keep the rest of the function the same.

hassy (approved):

PR #2211, opened 2023-10-12
feat(publish-metrics): increase trace visibility for OTel reporter

InesNi:
# What

Multiple improvements to OTel reporter tracing feature for the HTTP engine:

 ### Before:
   - Each request is its own root span and a separate trace with no information on which scenario it belongs to
   - No insight on individual request phases except for `responseTimeMs` attribute
   - Request spans have `http_request_started` and `http_request_ended` events
   - Errors are set as a `status` on spans
   
Individual trace:
Screenshot 2023-10-12 at 20 43 22

  
  ### After:
   - Scenarios are the root spans with all requests that are executed as a part of each scenario nested inside them as their child spans
   - Each request is broken down further into the following child spans (they represent the request phases respectively as they are recorded in the `got`  [`response.timings.phases`](https://github.com/sindresorhus/got/blob/main/source/core/response.ts) object): 
     - `dns_lookup` , 
     - `tcp_handshake` ,
     - `tls_negotiation` ,
     - `request` ,
     - `first_byte` ,
     - `download`
   - The `http_request_started` and `http_request_ended` events have been removed due to the implementation of the request phases spans
   - Errors that occur during execution (not the ones set due to response code being 400 or more) are also recorded with `span.setException()` not just as a span `status`,
   - virtual user id is set as an attribute on all request spans as `vu.uuid`
   - `responseTimeMs` attribute changed to `response.time.ms` for consistency
  
Individual trace:
Screenshot 2023-10-12 at 20 49 40

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4620,6 +4620,17 @@
         "node": ">=14"
       }
     },
+    "node_modules/@opentelemetry/context-async-hooks": {
+      "version": "1.17.1",
+      "resolved": "https://registry.npmjs.org/@opentelemetry/context-async-hooks/-/context-async-hooks-1.17.1.tgz",
+      "integrity": "sha512-up5I+RiQEkGrVEHtbAtmRgS+ZOnFh3shaDNHqZPBlGy+O92auL6yMmjzYpSKmJOGWowvs3fhVHePa8Exb5iHUg==",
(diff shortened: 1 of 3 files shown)


hassy:
This looks great, lgtm

PR #2212, opened 2023-10-13
fix(cloud): wait for unprocessed logs before shutdown

bernardobridge:
## Why

As logs take time to be sent as an API call, and there is also an onShutdown method that takes care of console logs, sometimes there isn't enough time for the logs to be sent to the API before Cloud's own onShutdown has already started processing.

We now wait for the unprocessed logs in the `onShutdown` before starting the actual cloud shutdown.

## Testing

Tested in Cloud with tests that were dropping logs (usually short tests had almost no logs, and longer tests tended to drop the Summary report). Problem went away.

You can also see this working by enabling `DEBUG=cloud`:

### Before
Notice how `logLines event` gets called after the Run URL has been printed, which is the last thing to be done `onShutdown` - those logs would be lost.
Screenshot 2023-10-13 at 09 51 11

### After
waiting on unprocessed logs -> `textlog sent event` -> all the cleanup hooks run -> url is printed
Screenshot 2023-10-13 at 09 50 02

### DIFF
--- a/packages/artillery/lib/platform/cloud/cloud.js
+++ b/packages/artillery/lib/platform/cloud/cloud.js
@@ -6,6 +6,8 @@
 
 const debug = require('debug')('cloud');
 const request = require('got');
+const awaitOnEE = require('../../util/await-on-ee');
+const sleep = require('../../util/sleep');
 const util = require('node:util');
 
 class ArtilleryCloudPlugin {
@@ -30,6 +32,7 @@ class ArtilleryCloudPlugin {
     this.defaultHeaders = {
       'x-auth-token': this.apiKey
     };
+    this.unprocessedLogs = [];
 
     let testEndInfo = {};
     global.artillery.globalEvents.on('test:init', async (testInfo) => {
@@ -86,6 +89,7 @@ class ArtilleryCloudPlugin {
 
     global.artillery.globalEvents.on('logLines', async (lines, ts) => {
       debug('logLines event', ts);
+      this.unprocessedLogs.push({ lines, ts });
 
       let text = '';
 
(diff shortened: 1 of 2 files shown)


packages/artillery/lib/util/await-on-ee.js
```diff
@@ -1,19 +1,21 @@
 const sleep = require('./sleep');
 
-async function awaitOnEE(ee, message, pollMs = 1000) {
+async function awaitOnEE(ee, message, pollMs = 1000, maxWaitMs = Infinity) {
```
bernardobridge, line 3: change made here to allow breaking the while loop. kept it backward compatible by introducing the default, so the current behaviour is not changed for other consumers of the function

packages/artillery/lib/platform/cloud/cloud.js
```diff
@@ -127,6 +137,18 @@ class ArtilleryCloudPlugin {
     global.artillery.ext({
       ext: 'onShutdown',
       method: async (opts) => {
+        const waitTimeInMs = 11000; //at most wait more than 10 seconds (report interval is 10s)
```
hassy, line 140: Would on `onShutdown` ever get called before all VUs are finished and reports are received? How likely is it that Artillery is now going to take 10s+ to stop every time?

bernardobridge: Actually @hassy , now that I think about it, do we still need that `await sleep(5 * 1000);` there, at least to such a large number? Was it put there to avoid this precise problem? Perhaps with this implementation it can also be reduced, or is no longer needed?

bernardobridge: Actually @hassy I've been thinking about this a little bit more - since you implemented it, do you remember if there's a reason other than this exact one why that `await sleep(5 * 1000)` is there? 

If not, then I think perhaps we might be able to remove it (or greatly reduce it), as this new mechanism now waits for the actual events to be sent. So if there wasn't another reason, it might be an opportunity to revisit it?

PR #2233, opened 2023-10-19
fix(docker): remove playwright browsers from image

bernardobridge:
## Description

Related to https://github.com/artilleryio/artillery/pull/2230, where this was removed. The playwright image (`mcr.microsoft.com/playwright:v1.39.0`) actually still contains all the browsers, regardless of what we install in the package, so we still need to manually remove them.


joshuscurtis:
Hi - I am seeing the following error:

looks like it could be related to this change?

`worker error, id: 1 browserType.launch: Executable doesn't exist at /ms-playwright/chromium-1076/chrome-linux/chrome`

PR #2274, opened 2023-11-02
feat(artillery): collect per url metrics by default

InesNi:
# Description

## What
Collecting per-url metrics by default by loading the `metrics-by-endpoint` plugin by default behind the scenes.

## How

- Creating `addDefaultPlugins` function that takes the final version of the script and adds the  default plugins and their config if they are not already listed on the script. (For now we only add `metrics-by-endpoint` by default but the function is created with extensibility in mind.)
The `metrics-by-endpoint` plugin  is set on the script with the two config settings: 
  - `suppressOutput` - a newly created setting, used to exclude plugin created metrics from console
  - `stripQueryString` - existing setting, removes query string from endpoint

- `addDefaultPlugins` is then called in the `prepareTestExecutionPlan` function as a last modification to the script.
- The final script is then passed to the `ConsoleReporter` where we use it to check for `suppressOutput` along with the default metric prefix to filter out and exclude plugin created metrics when reporting to the console.
It is worth noting that if `suppressOutput` setting is ever exposed to the user we would have to check for the `metricsPrefix` setting here as well. 

## Testing

A `'Loads metrics-by-endpoint plugin by default, with output supressed'` test is created to test the new functionality, and it has been tested manually.

# Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/cmds/run.js
+++ b/packages/artillery/lib/cmds/run.js
@@ -5,6 +5,7 @@ const {
   parseScript,
   addOverrides,
   addVariables,
+  addDefaultPlugins,
   resolveConfigTemplates,
   checkConfig
 } = require('../../util');
@@ -265,7 +266,8 @@ RunCommand.runCommandImplementation = async function (flags, argv, args) {
 
     // TODO: Wire up workerLog or something like that
     const consoleReporter = createConsoleReporter(launcher.events, {
-      quiet: flags.quiet || false
+      quiet: flags.quiet || false,
+      script
     });
 
     let reporters = [consoleReporter];
@@ -492,7 +494,9 @@ async function prepareTestExecutionPlan(inputFiles, flags, args) {
   }
 
   script5.config.statsInterval = script5.config.statsInterval || 30;
-  return script5;
+
+  const script6 = addDefaultPlugins(script5);
+  return script6;
 }
 
 async function readPayload(script) {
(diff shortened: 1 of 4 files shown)


packages/artillery/lib/cmds/run.js
```diff
@@ -263,9 +264,11 @@ RunCommand.runCommandImplementation = async function (flags, argv, args) {
     );
     let intermediates = [];
 
+    const suppressMetrics = getPluginMetricsToSuppress(script) || []
```
hassy, line 267:
```suggestion
    const suppressMetrics = getPluginMetricsToSuppress(script);
```

hassy: The function should always return a list imo.

packages/artillery/lib/cmds/run.js
```diff
@@ -684,4 +689,18 @@ function getLogFilename(output, nameFormat) {
   return logfile;
 }
 
+function getPluginMetricsToSuppress(script){
```
hassy, line 692: Looks like `prettier` didn't run here?

packages/artillery/lib/console-reporter.js
```diff
@@ -29,6 +29,7 @@ function ConsoleReporter(opts) {
   this.outputFormat = opts.outputFormat || process.env.OUTPUT_FORMAT || 'new';
 
   this.quiet = opts.quiet;
+  this.metricsToSuppress = opts.suppressMetrics;
```
hassy, line 32: Nit but should probably just use the same name for external config and internally? (`metricsToSuppress` or `suppressMetrics`)

PR #2284, opened 2023-11-08
fix(ensure): improve handling of metric names with special chars

bernardobridge:
## Description

This PR is a rework of https://github.com/artilleryio/artillery/pull/2180.

Originally reported here: https://github.com/artilleryio/artillery/discussions/2166#discussioncomment-7110325, ensure is currently broken for most Playwright test assertions that include urls. The problem is more general though - any metric that includes special characters can hit this (such as custom metrics, created by name and metrics by endpoint).

### Approach

As it's hard to predict what patterns can fall into this (as it depends on how the package filtrex is handling it), rather than going with an approach of replacing specific patterns, we hash all metricNames in the expression and variables, and use that with filtrex.

Unlike the previous PR, the source of truth for metric names now comes from the list of metrics themselves. As long as the metric exists, it will be replaced and hashed in the expression. This avoids us having complicated regexes to do pattern recognition in the expression defined by the user.

### Testing
This has been thoroughly tested, as you can see from the added unit tests and acceptance tests, including with all sorts of weird characters and expressions. If you can think of another edge case, please let me know. As you can see, it's now even working against Artillery Cloud:

Screenshot 2023-11-08 at 17 55 51

### Limitations
- I only encountered one limitation, and it's a bit of an edge case: if the metric name doesn't exist (e.g. a web vital that never got emitted) AND it has characters that filtrex doesn't like, it will error [in these try catches](https://github.com/artilleryio/artillery/blob/7f0201829853af4e084e52f6d088f81d6ba02ce3/packages/artillery-plugin-ensure/index.js#L140-L142)).
- We may want to change that from `global.artillery.log` to only show errors with `debug`.

Still, this seemed acceptable to me, as it's still much better for most cases, and not any worse for that one particular edge case. This is also IMO the simplest solution that avoids tanking more time into filtrex or other parsing solutions.

## Pre-merge checklist

- [ ] Does this require an update to the docs? _We could add a warning to the docs about non-existing metrics. Other than that, no._
- [x] Does this require a changelog entry? _Yes_

### DIFF
--- a/packages/artillery-plugin-ensure/index.js
+++ b/packages/artillery-plugin-ensure/index.js
@@ -7,6 +7,11 @@
 const debug = require('debug')('plugin:ensure');
 const filtrex = require('filtrex').compileExpression;
 const chalk = require('chalk');
+const {
+  hashString,
+  replaceMetricsWithHashes,
+  getHashedVarToValueMap
+} = require('./utils');
 
 class EnsurePlugin {
   constructor(script, events) {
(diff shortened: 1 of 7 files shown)


packages/artillery-plugin-ensure/index.js
```diff
@@ -65,29 +70,49 @@ class EnsurePlugin {
...
+                global.artillery.suggestedExitCode = 1;
+              }
+            } else {
+              global.artillery.log(`${chalk.green('ok')}: ${check.original}`);
             }
-          } else {
-            global.artillery.log(`${chalk.green('ok')}: ${check.original}`);
-          }
-        });
+          });
```
bernardobridge, line 87: You can disregard this part, as it is unrelated to the change, happened due to running lint/prettier here: https://github.com/artilleryio/artillery/pull/2284/commits/717d290e167f61b335a189623560e949e2c3e774

(this file was failing linting due to radix missing in a parameter, so prettier hadn't actually run on it)

hassy (approved):
lgtm

PR #2285, opened 2023-11-08
fix(publish-metrics): fix vu.uuid and improve attribute handling

InesNi:
# Description

- Fix vu.uuid attribute setting  

- Set attributes available in the beforeRequest hook before the request so they are available if an error occurs before the response

- Use `SemanticAttributes` object when setting attributes on spans. This ensures consistency across different instruments and systems.
- Add `HTTP_REQUEST_CONTENT_LENGTH`, `HTTP_USER_AGENT` and `HTTP_FLAVOR` semantic attributes to the request spans

- Parse URL to remove possible authentication details from the URL as required by convention:
Screenshot 2023-11-17 at 14 25 32

- Add span error message along with the code for response status of 400 or more

# Pre-merge checklist

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
+++ b/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
@@ -132,6 +132,8 @@ class OTelReporter {
         'trace'
       );
       this.tracing = true;
+      this.pendingRequestSpans = 0;
+      this.pendingScenarioSpans = 0;
 
       this.configureTrace(this.traceConfig);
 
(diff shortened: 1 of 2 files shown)


bernardobridge (approved):

bernardobridge (changes requested):
~is there a reason to remove this? I think we should still set the message if available, no?~

Nevermind, misread the diff. All good 👍

packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -459,23 +482,22 @@ class OTelReporter {
...
       });
 
       if (res.statusCode >= 400) {
-        span.setStatus({ code: SpanStatusCode.ERROR });
-      }
-      if (this.traceConfig?.attributes) {
-        span.setAttributes(this.traceConfig.attributes);
+        span.setStatus({
+          code: SpanStatusCode.ERROR,
+          message: res.statusMessage
```
bernardobridge, line 497: is there a reason to remove this? I think we should still set the message if available, no?

bernardobridge (changes requested):

bernardobridge (approved):

PR #2293, opened 2023-11-15
feat(artillery): expose graceful shutdown and its options globally

bernardobridge:
## Description

This PR is a prerequisite for some more changes where we will need to call `gracefulShutdown` from within a plugin. This makes that possible.

Additionally, it exposes the `gracefulShutdown` options to `beforeExit` hooks. This makes it possible for specific plugins to decide if they want to be run when exiting early. 

In the case of `ensure` plugin, it makes sense not to run it (at least for now), since it will always fail, not because of not running checks, but due to the `stats` final object not existing at that point.

## Pre-merge checklist

- [ ] Does this require an update to the docs? I think `beforeExit` and `onShutdown` hooks need to be properly documented first - then this can go there

### DIFF
--- a/packages/artillery-plugin-apdex/index.js
+++ b/packages/artillery-plugin-apdex/index.js
@@ -66,6 +66,13 @@ class ApdexPlugin {
           return;
         }
 
+        if (testInfo.earlyStop) {
+          console.log(
+            'WARNING: Test run exited early. Skipping apdex calculation.'
+          );
+          return;
+        }
+
         const s = testInfo.report.counters[METRICS.satisfied] || 0;
         const t = testInfo.report.counters[METRICS.tolerated] || 0;
         const f = testInfo.report.counters[METRICS.frustrated] || 0;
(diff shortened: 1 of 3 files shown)


packages/artillery-plugin-apdex/index.js
```diff
@@ -66,6 +66,13 @@ class ApdexPlugin {
           return;
         }
 
+        if (testInfo.earlyStop) {
```
hassy, line 69: I don't think we want to change current behavior. There is no clear reason to **not** always run these checks, so I'm hesitant to change existing behavior.

hassy (changes requested):
We should not change existing behavior of the two plugins.

packages/artillery-plugin-apdex/index.js
bernardobridge: okay - reverted. we'll address making cancellation more obvious at another point

hassy (approved):

PR #2297, opened 2023-11-17
refactor: import @artilleryio/platform-fargate

hassy:
This moves existing code from the external package into this repo mostly as-is. The code has been linted / prettified with the config in the repo, and some of the more obvious unnecessary Artillery Pro-isms have been removed.

This works end-to-end without the external dependency now.

Docker image building + publishing pipeline, and existing e2e test suite for Fargate will be moved over as separate PRs.


hassy:
Merging this to make testing easier with a canary release.

PR #2298, opened 2023-11-17
fix(publish-metrics): use onError hook for error handling

InesNi:
# Description

Currently, in the case of errors that happen before the request, the spans are being inconsistently reported, and the errors are not always set on the spans they belong to.

This PR implements handling the errors in the existing `onError` hook. In order to accomplish this the change had to be made from two sides:

- In the OpenTelemetry reporter, the before request error handling has been moved to the new hook function.
- In the HTTP engine, the existing `runOnErrorHooks` has been implemented in the `catch` block of the request execution

## Notes
- The 'error' [event listener callback](https://github.com/artilleryio/artillery/blob/main/packages/core/lib/engine_http.js#L693) does not seem to catch errors, they all propagate to the `catch` and are caught there. This will need further investigation
- Calling the `runOnErrorHooks` in the `catch`  callback activated the `onError` [hook function](https://github.com/artilleryio/artillery/blob/main/packages/artillery-plugin-expect/index.js#L79) in the  `expect` plugin which did not run before (due to the issue mentioned in previous note) This will need to be addressed in a separate PR

# Pre-merge checklist


packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -385,7 +394,10 @@ class OTelReporter {
   endScenarioSpan(engine) {
     return function (userContext, ee, next) {
       const span = userContext.vars[`__${engine}ScenarioSpan`];
-      span.end(Date.now());
+      if (!span._ended) {
```
bernardobridge, line 397: `._ended` seems to be a private option, so maybe best not to rely on it if we can avoid it. 

Looking at the code it seems there's also a `this.endTime` that they set (https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/Span.ts#L78). Maybe we could use that instead?

bernardobridge (changes requested):
Just a small change, if it works. Otherwise I think using `_ended` for now should be ok.

bernardobridge (approved):

PR #2301, opened 2023-11-18
feat: CPU and memory configuration on Fargate

hassy:
- Add `--cpu` and `--memory` flags for convenience
- Update the table of valid CPU/memory configurations

Fixes #2280.

## Pre-merge checklist

- [ ] Does this require an update to the docs? Yes, todo.
- [ ] Does this require a changelog entry? Yes, todo.

### DIFF
--- a/packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
+++ b/packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
@@ -347,6 +347,39 @@ async function tryRunCluster(scriptPath, options, artilleryReporter) {
       process.exit(1);
     }
     options.launchConfig = launchConfig;
+  } else {
+    options.launchConfig = {};
+  }
+
+  if (options.cpu) {
+    const n = Number(options.cpu);
+    if (isNaN(n)) {
+      artillery.log('The value of --cpu must be a number');
+      process.exit(1);
+    }
+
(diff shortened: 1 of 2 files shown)


packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
```diff
@@ -347,6 +347,38 @@ async function tryRunCluster(scriptPath, options, artilleryReporter) {
...
+  if (options.cpu) {
+    const n = Number(options.cpu);
+    if (isNaN(n)) {
+      artillery.log('The value of --cpu must be a number');
+      process.exit(1);
+    }
+
+    // Allow specifying 16 vCPU as either "16" or "16384". The actual value is
+    // validated later.
+    if (n <= 16) {
```
bernardobridge, line 363: nit: can we put the number 16 as a const somewhere, rather than being a magic number? There's a comment, but still took me a sec to think "oh yeah, that's the max vCPU value currently available"

hassy: yes good shout

hassy:
```suggestion
    const MAX_VCPUS = 16;
    if (n <= MAX_VCPUS) {
```

PR #2315, opened 2023-11-23
refactor(expect): exclude failures from errors logged

InesNi:
# Description

## What
Exclude failed expectation errors (present when `reportFailuresAsErrors` is set to true) from errors being logged to console with `expectationsPluginOnError` function.

## Why
Currently expect plugins `onError` hook function - `expectationsPluginOnError` logs the `'Error: {error message}'` string ( 'Error' being logged in red chalk), on every error. As `'not ok'` is already logged in red chalk for failed expectations, having additional 'Error' logged in red chalk is unnecessary. 

## Notes
- The `expectationsPluginOnError` function was dormant until #2298 due to `onError` hook not functioning properly 

# Pre-merge checklist
 - [] Does this require an update to the docs?

### DIFF
--- a/packages/artillery-plugin-expect/index.js
+++ b/packages/artillery-plugin-expect/index.js
@@ -83,7 +83,7 @@ function expectationsPluginOnError(
   events,
   done
 ) {
-  if (scenarioErr.message.startsWith('Failed expectations for request')) {
+  if (scenarioErr instanceof FailedExpectationError) {
     return done();
   }
   if (userContext.expectationsPlugin.formatter === 'json') {
@@ -186,7 +186,9 @@ function expectationsPluginCheckExpectations(
         ? req.name
         : req.url;
     return done(
-      new Error(`Failed expectations for request ${filteredRequestName}`)
+      new FailedExpectationError(
+        `Failed expectations for request ${filteredRequestName}`
+      )
     );
   }
 
@@ -211,5 +213,12 @@ function maybeParseBody(res) {
     return body;
   } else {
     return res.body;
+  }
+}
+
+class FailedExpectationError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = 'FailedExpectationError';
   }
 }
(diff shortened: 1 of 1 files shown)


packages/artillery-plugin-expect/index.js
```diff
@@ -83,10 +83,13 @@ function expectationsPluginOnError(
...
-  if (userContext.expectationsPlugin.outputFormat === 'json') {
-    artillery.log(JSON.stringify({ ok: false, error: scenarioErr.message }));
+  if (scenarioErr.message.startsWith('Failed expectations for request')) {
+    return done();
+  }
+  if (userContext.expectationsPlugin.formatter === 'json') {
+    artillery.log(JSON.stringify({ error: scenarioErr.message }));
   } else {
-    artillery.log(`${chalk.red('Error:')} ${scenarioErr.message}`);
+    artillery.log(`\n${chalk.red('Error:')} ${scenarioErr.message}\n`);
```
hassy, line 92:
```suggestion
    artillery.log(`${chalk.red('Error:')} ${scenarioErr.message}`);
```

Suggest we revert the formatting change as well here, as it's unrelated to the fix.

packages/artillery-plugin-expect/index.js
```diff
@@ -83,10 +83,13 @@ function expectationsPluginOnError(
   events,
   done
 ) {
-  if (userContext.expectationsPlugin.outputFormat === 'json') {
-    artillery.log(JSON.stringify({ ok: false, error: scenarioErr.message }));
+  if (scenarioErr.message.startsWith('Failed expectations for request')) {
+    return done();
+  }
+  if (userContext.expectationsPlugin.formatter === 'json') {
+    artillery.log(JSON.stringify({ error: scenarioErr.message }));
```
hassy, line 90:
```suggestion
    artillery.log(JSON.stringify({ ok: false, error: scenarioErr.message }));
```

Best not to remove fields someone might rely on without a good reason.

packages/artillery-plugin-expect/index.js
```diff
@@ -83,10 +83,13 @@ function expectationsPluginOnError(
   events,
   done
 ) {
-  if (userContext.expectationsPlugin.outputFormat === 'json') {
-    artillery.log(JSON.stringify({ ok: false, error: scenarioErr.message }));
+  if (scenarioErr.message.startsWith('Failed expectations for request')) {
```
hassy, line 86: This will work for now but probably better to handle this without relying on the error message, e.g. with a custom error class.

PR #2316, opened 2023-11-23
fix(cloud): set status interval after testrun init event

bernardobridge:
## Description

Noticed a couple of errors in the Artillery Cloud backend due to calling `/status` endpoint before test run exists in the DB. Since the stop test button only appears after the init event has completed, then we only need to start polling for `/status` at that point too.

## Pre-merge checklist

- [ ] Does this require a changelog entry? No, it's a fix for an unreleased feature


hassy:
:shipit:

PR #2333, opened 2023-11-28
Bernardobridge/art 1482 cli test run end changes

bernardobridge:
## Description

Improves performance of the Cloud `testrun:end`, which makes the CLI end the test faster.

## Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/cmds/run.js
+++ b/packages/artillery/lib/cmds/run.js
@@ -407,7 +407,7 @@ RunCommand.runCommandImplementation = async function (flags, argv, args) {
       const ps2 = [];
       for (const e of global.artillery.extensionEvents) {
         if (e.ext === 'onShutdown') {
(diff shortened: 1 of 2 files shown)


hassy:
Is there a way we could do this without passing another argument to `onShutdown` handlers? Would be best to keep the number of arguments those to a minimum as we don't want to encourage using them over `beforeExit` handlers.

bernardobridge:
>Is there a way we could do this without passing another argument to onShutdown handlers? Would be best to keep the number of arguments those to a minimum as we don't want to encourage using them over beforeExit handlers.

Nice shout @hassy . Turns out we already had a mechanism to pass data from `beforeExit` to `onShutdown` in the cloud platform, so I can just leverage that: https://github.com/artilleryio/artillery/pull/2333/commits/c218de8a5e8be57e0995051a5981e86f2f3465fc

PR #2340, opened 2023-11-30
feat(publish-metrics): send metrics to prometheus on done or stats event

pmtkh:
## Description

Gives possibility to send metrics to Prometheus only summary report (on done event). This is configurable in test config using property `sendMetricsAtSummary` with boolean value.

**Reason:**
Currently metrics are send only at stats event which brings two issues:
- you cannot define mare than one phase - not all metrics are send to Prometheus
- summary metrics are not recalculated when send multiple metrics (based on `stats` event) - in Prometheus you see only latest metric send by Artillery.io.

## Pre-merge checklist

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/prometheus.js
+++ b/packages/artillery-plugin-publish-metrics/lib/prometheus.js
@@ -107,8 +107,9 @@ class PrometheusReporter {
 
   sendMetrics(config, events) {
     let that = this;
+    const eventType = config.sendMetricsAtSummary ? 'done' : 'stats';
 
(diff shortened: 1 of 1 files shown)


hassy:
thank you for the PR @pmtkh! I'd like to understand the use case a little bit better. `publish-metrics` sends metrics continuously by design, so that the progression of a load test can be tracked across time. By only sending a summary at the end you'd lose that and the granularity of metrics.

pmtkh:
@hassy is there any plan to merge changes prepared within this PR?

Issue #1873, opened 2023-03-16
Question: Is it possible to add custom information to the report?


eliezercazares:
I.e. I am running functional testing and I would like the request Headers, URL/URI and Payload/Body to be included in the report.

I don't mind having a .yml per scenario and one report per scenario, I can merge them. It would be more useful for me in this particular case to be able to include such information on the report.

I would also love to add things like:

`- report: "In the response JSON there were {{ amountNodes }} nodes, from {{ totalNodes }} possible nodes"`

Like, if I managed some validations and operations in a processor custom.js method and stored the results in a variable, then I would love to print that to the report.

Zacaria:
HI !
+ 1 I'd like to add a message, and the config which was involved in the benchmark

SpencerWightman:
I would also like to include a message in the report. I'll take a crack at implementing it.

Issue #1005, opened 2021-04-26
Add ability to mix multiple engines in the same scenario


hassy:
It should be possible to use multiple engine in the same scenario, to for example:
- Send a POST request to an HTTP endpoint to get an auth token
- Use that auth token to connect to a WS endpoint
- Send and receive some data from the WS server
- Use some of that data to post a message to a Kafka topic

dav1app:
I have exactly this flow on my tests. Is there anything up to this feat over the `dev`?

stanyq4:
We are looking to cover a similar load test, any updates?

hassy:
Not yet

qiulang:
Hi I am a bit confused with this issue. Here https://www.artillery.io/blog/load-testing-socketio-with-artillery said 

> You can also mix HTTP and Socket.io actions in the same scenario.

But this issue is still open so can it or can't it ?

Thanks.

hassy:
@qiulang this issue is about allowing any engines to be combined. Socket.io has custom logic to integrate with the HTTP engine, so you can use both Socket.io and HTTP actions in the same scenarioo.

liyoung1874:
@hassy How to use http protocol and wss protocol calls in one flow ?

GP-QA:
Hi, is it possible to use different engines per scenario? I would like to use Playwright in a before hook to login and grab the 0auth 2.0 bearer token and re-use it in other HTTP scenarios. It is not possible in my web application to generate it only via a API call.

tobyspark:
Flagging that this remains a huge hole from my POV. I wish to simulate a page with socket-based sync. i.e. `http` for `get`, then `ws` for `connect` etc.

Further, Artillery is broken for the workarounds. The `ws` engine just seems incompatible with processors. Using the `http` engine I see the expected logging from my `beforeScenario` function that sets context variables for the websocket to connect with. Set to the `ws` engine I just see websocket connection errors, with the variables used in the `connect` URL `undefined`.

PR #2348, opened 2023-12-04
Issue 1873 custom message metric feature

SpencerWightman:
## What

I added a `'custom message'` metric to the existing [Custom Metrics API](https://www.artillery.io/docs/reference/extension-apis#custom-metrics-api) that catches a processor file's `'custom message'` emit. It displays each custom message from the time period in the period report on separate lines. This was done to address issue #1873.

Screenshot 2023-12-04 at 3 12 28 PM

## Why

In some cases users want to add custom messages/data to the period reports. Two people in the issue thread plus myself expressed a desire for the feature. Furthermore, @bernardobridge added the `enhancement` label to the issue.

## How

In summary, the `'custom message'` metric was added by integrating it with the current Custom Metrics API implementation flow, primarily in `SSMS.js`, `runner.js`, and `console-reporter.js`. It used the other custom metric implementations as a template.

## Testing

I added a test to `artillery/test/unit/ssms.basic.test.js` for the `customMessage` metric, alongside the tests for the other metrics. All the tests in `ssms.basic.test.js` passed after my addition. All the tests in `artillery/test/runner.sh` passed except for the 2nd and 3rd. However, those tests also failed when run on branch main. All of the `./run.sh` tests in `packages/core` passed.


hassy:
Thanks for taking the time @SpencerWightman! I'd love to understand the exact use case of something like this, e.g. what type of information would be particularly useful to have attached to a specific time period.

SpencerWightman:
No problem at all @hassy. Artillery is awesome and I’m happy to contribute!

At a high level, the use case is when a developer extracts scenario logic to a JS file and wants to log data from that file.

Extracting scenario logic can occur because a developer finds it necessary ([mixing engines](https://github.com/artilleryio/artillery/issues/1005) in the same scenario, for example) or because they find implementing a complex scenario in JS preferable. Logged data can include response bodies, headers, WebSocket emitted data, UUIDs, or any other data, either raw or after helper function processing.

The `customMessage` feature, implemented through the Custom Metrics API, is simpler and more flexible than storing `function` action values with `context.vars` and then using `- log: "{{ name }}"` in the YAML:

1. Emitting within the JS file allows developers to easily create as many messages as they want, exactly where they want in the code, per virtual user. Example: a value changes through an iterative process or based on a conditional, and the developer wants to log the value selectively, perhaps along with a precise timestamp or other ancillary data. This would be cumbersome or impossible with the `context.vars` system because 1) `- log: "{{ name }}"` always logs a value and 2) `context.vars.name` is always the latest `name` value.

2. Using `context.vars` in JS to log values from the YAML file, then adding logs to the YAML, is less straightforward than having logging logic in the JS file along with the rest of the scenario code (especially if there are many values to log). The `customMessage` feature requires less code, and the code is in one file.

3. Messages logged through the `context.vars` system are not within a period report and are more difficult to parse in the overall output as a result. Logging within a period report is also desirable for clear alignment with testing periods.

In summary, the option to output custom period report data from a JS file means more control and convenience in complex testing situations.

I’m happy to elaborate if any of that is too vague!

SpencerWightman:
Hi @hassy, is there anything I can do to move this forward? This feature adds intuitive custom messaging for developers. The code integrates smoothly into the Custom Metrics API and has a test.

hassy:
Hi @SpencerWightman, apologies for the delay. Agree that a way to associate a text log with a specific time period may be useful, but the implementation should not be via the metrics API. Using a `counter` metric as already available would provide essentially the same functionality as the implementation in this PR. We'll look at supporting time-specific logs at some point, but it's not a priority right now. Thank you for taking the time, and apologies again for the delay.

PR #2360, opened 2023-12-08
feat(cli): add typescript support by bundling processor with esbuild

bernardobridge:
## Description

Adds support for typescript processors in run and run-fargate.

Some of the decisions that were explored and ended up being made:
- Disable for now in `aws:lambda` due to an issue with bundle sizes. Will be explored again later. As this is primarily for Playwright Typescript support, and Playwright does not run in Lambda, this is not urgent;
- Using `esbuild-wasm` as it is cross-platform compatible. Regular `esbuild` also works and installs the platform specific binary, but this is more future-proof for when we do provide Lambda support. This can easily be changed;
- For now using `global.artillery.hasTypescriptProcessor`. This can be changed a little bit later by refactoring some of the logic in `run`, but right now it felt unnecessary to tackle immediately.

## Pre-merge checklist

- [ ] Does this require an update to the docs? Yes!
- [ ] Does this require a changelog entry? Yes!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8854,6 +8854,17 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/esbuild-wasm": {
+      "version": "0.19.8",
+      "resolved": "https://registry.npmjs.org/esbuild-wasm/-/esbuild-wasm-0.19.8.tgz",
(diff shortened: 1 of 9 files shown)


bernardobridge:
Some action items to do here:

- [x] Not allow running in AWS Lambda for now (due to package size issues)
- [ ] Ensure works with configs in different folders
- [ ] Error handling
- [ ] Revisit some of the esbuild configs
- [ ] Create some tests

PR #2366, opened 2023-12-14
feat(fargate): include datadog agent in worker image

InesNi:
# Description

Since Datadog is different to other vendors in first-class support for OTeL, and it's not as simple as just an API endpoint and headers (data via OTLP can only be ingested through the DD agent or the OTel collector), we decided to provide a solution that would make it easier to send data to Datadog via OTel reporter when running on Fargate.

In order to accomplish this the Datadog agent is bundled with the worker image. 
Installation is being done using [DD's install script](https://app.datadoghq.com/account/settings/agent/latest?platform=ubuntu), as that is what the DD documentation shows. The Ubuntu platform script was chosen, as the Docker one seems to be aimed at starting this as a sidecar container, and Ubuntu is what the Playwright image uses.

The agent can not be installed without providing an API key so we set a dummy one on installation and then override it before starting the agent.

## Configuration
- In order to enable the Datadog agent in Fargate the user needs to set the `DD_API_KEY` environment variable in a `dotenv` file, and provide the file path in the run command with the `--dotenv` flag as follows:

  **artillery run-fargate < path to your test script > --dotenv <path to your `dotenv` file**

In this case no `endpoint` or  `headers` need to be provided to the OpenTelemetry reporter, as it will use the default signal specific endpoints. [Reference](https://opentelemetry.io/docs/concepts/sdk-configuration/otlp-exporter-configuration/#otel_exporter_otlp_traces_endpoint)

**Example config in test script:**
```yaml
  plugins:
    publish-metrics:
      - type: open-telemetry
        serviceName: your_service_name
        traces: {}
```

**Example command:**

```stout
artillery run-fargate scenario.yml --dotenv ./.env
```

No modifications were needed to the OpenTelemetry reporter itself.

## Notes
- **The image size has a significant increase (40%, 1GB -> 1.4GB) due to the agent size**, which seems to be a known [issue](https://github.com/DataDog/datadog-agent/issues/3542). 
However, the startup time does not seem to be significantly affected by this so it was decided to proceed with this implementation and optimise in future releases

- Instructions for sending the data over the OpenTelemetry reporter to Datadog locally are provided [here](https://artillery.notion.site/Playwright-tracing-with-Datadog-local-set-up-bb23b137915b48aa80854afbc5a8b726) and will be added to the docs.

# Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/platform/aws-ecs/worker/dd-agent-setup.sh
+++ b/packages/artillery/lib/platform/aws-ecs/worker/dd-agent-setup.sh
@@ -21,6 +21,7 @@ start_and_configure_dd_agent() {
 
     # Add otlp config
     yq -i ".otlp_config.receiver.protocols.http.endpoint = \"localhost:4318\"" "$yaml_file"
+    yq -i ".otlp_config.receiver.protocols.grpc.endpoint = \"localhost:4317\"" "$yaml_file"
 
     # Add apm_config trace_buffer
     yq -i ".apm_config.trace_buffer = 100" "$yaml_file"
--- a/packages/artillery/lib/platform/aws-ecs/worker/dd-agent-setup.sh
+++ b/packages/artillery/lib/platform/aws-ecs/worker/dd-agent-setup.sh
@@ -2,6 +2,7 @@
 start_and_configure_dd_agent() {
     # Check if DD_API_KEY is set in the environment
     if [ -z "${DD_API_KEY-}" ]; then
+        echo "DD_API_KEY not set. Not running Datadog Agent."
         return 0
     fi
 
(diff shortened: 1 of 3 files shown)


packages/artillery/lib/platform/aws-ecs/worker/dd-agent-setup.sh
```diff
@@ -0,0 +1,35 @@
+
+start_and_configure_dd_agent() {
```
hassy, line 2: Could we just configure the agent through env vars instead of using a config file that we have to modify.

InesNi: Hi @hassy 👋🏻 

Maybe I missed something, but I've tried everything I could think of and unfortunately I don't believe that's possible. In another branch I tried setting `DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_HTTP_ENDPOINT` and `DD_APM_TRACE_BUFFER` (the same ones set for local run through Docker), as well as some other variations, and it doesn't work in either case:
- installing (in docker build) and separately starting
- or install+start directly in the loadgen-worker script. 

The config just doesn't get picked up (I even checked the `datadog.yaml` file and it doesn't have the configuration options set there).

This is what I meant before when I said I thought that for different platforms their install script handles environment variables for certain features (otlp, apm) differently. Works with their Docker image, but doesn't seem to work with host install like this. 

Maybe we can revisit this when we try using a sidecar?

InesNi:
Closed in favour of https://github.com/artilleryio/artillery/pull/2381 (sidecar container)

PR #2396, opened 2024-01-04
fix(cli): allow beforeafter validation schema to run engines without flow

bernardobridge:
## Description

In theory any `engine` can run as a `before/after` hook. However, if the engine doesn't include `flow`, it will currently fail. This PR fixes that, and also introduces an error message when `before.engine` is configured without the corresponding `config.engine` (similar to https://github.com/artilleryio/artillery/pull/2101).

Having this will enable e.g. `playwright` scenarios to be run as a `before`.

Note: I will fix the VS code schema separately in another PR.

## Pre-merge checklist

- [ ] Does this require an update to the docs? Yes, maybe clarify usage of before/after with other engines


packages/artillery/lib/util/validate-script.js
```diff
@@ -105,9 +105,13 @@ const scenarioItem = Joi.object({
 });
 
 const beforeAfterSchema = Joi.object({
```
bernardobridge, line 107: Previously the schema would look at `config.engines` to determine this. However, this is more complicated than just looking at the `engine` field within before/after itself:
- If it doesn't have it, it's http (default), so it needs flow.
- If it has it, if it's `socketio`, `ws` or `http`, it needs flow. Otherwise it does not require flow (e.g. custom engine or Playwright)

packages/core/lib/runner.js
```diff
@@ -493,6 +493,12 @@ function handleScriptHook(hook, script, hookEvents, contextVars = {}) {
 
     const name = script[hook].engine || 'http';
     const engine = engines.find((e) => e.__name === name);
+
```
bernardobridge, line 496: similar logic to https://github.com/artilleryio/artillery/pull/2101 for consistency

PR #2400, opened 2024-01-09
feat(cli): record publish-metrics reporters usage with posthog

InesNi:
# Description

Extending `sendTelemetry` function to send `publish-metrics` reporters used in order to get insights into which reporters are most popular with users.

# Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/cmds/run.js
+++ b/packages/artillery/lib/cmds/run.js
@@ -639,6 +639,30 @@ async function sendTelemetry(script, flags, extraProps) {
       }
     }
 
+    // publish-metrics reporters
+    if (script.config.plugins['publish-metrics']) {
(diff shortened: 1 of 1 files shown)


packages/artillery/lib/cmds/run.js
```diff
@@ -639,6 +639,30 @@ async function sendTelemetry(script, flags, extraProps) {
...
+        'newrelic',
+        'splunk',
+        'dynatrace',
+        'cloudwatch',
+        'honeycomb',
+        'mixpanel',
+        'prometheus'
+      ];
+
+      properties.officialReporters = script.config.plugins[
```
hassy, line 657:
```suggestion
      properties.officialMonitoringReporters = script.config.plugins[
```

PR #2406, opened 2024-01-10
ci: implement aws tests on prs

bernardobridge:
## Description

Allows to run tests for our AWS platforms (Lambda, Fargate) from a branch securely (even from forks). It does this by:
- Using `pull_request_target` so that the necessary secrets are available to the job;
- Checks for permissions using `actions-cool/check-user-permission@v2` as a first conditional step. If the permissions are not those of a collaborator, then it will not run any subsequent steps.
- Since `pull_request_target` only has the context of the base branch (`main`), we checkout the specific ref needed to ensure the correct (branch) code is available.
- A docker image is built against this code as the first job after `is-collaborator`, so Fargate can be tested. Then, the tests run.

*Note: As Github requires workflows to be on `main` before they run the first time, the new workflow will not run on this PR. For this reason, I only fixed the single Lambda test. A PR after this will port over the Fargate tests, but I've tested the permissions and they work (i.e. run on Fargate).*

*Note 2: this PR is best reviewed by commit, as the changes are isolated on each commit.*

## Pre-merge checklist


.github/workflows/run-aws-tests.yml
```diff
@@ -0,0 +1,71 @@
+name: Run AWS tests
+
+on:
+  pull_request_target:
+    branches: [main]
+    #opened, reopened and synchronize will cause the workflow to fail on forks due to permissions
+    #once labeled, that will then be overridden by the is-collaborator job
+    types: [opened, labeled, synchronize, reopened]
```
hassy, line 8: Do we need all of those triggers here if we're only using `labeled`?

bernardobridge: IMO yes, as it provides a good DX. Here's why:

## opened
Needs to be here so it will block the PR when opened by a `fork`. It will run, check permissions aren't good, and fail the workflow. That then requires one of us to come and label. Essentially it's what gives the "auto-block" feature we discussed.

## synchronize/reopened
The idea with these, especially `synchronize` is that a change in the branch will cause the tests to automatically rerun (without needing to relabel). 

Again, if the change was made from a non-collaborator, it will block the PR again and one of us will need to label. If it's a change from a collaborator, the tests should just run

hassy: okay makes sense!

PR #2421, opened 2024-01-16
fix(fargate): send taskArns of non-leader tasks to cloud

bernardobridge:
## Description

Only the arn of the leader task was being sent to cloud. This sends the arn of other created tasks too, as they're created.

## Pre-merge checklist


bernardobridge:
Merging this - running into known issue with Fargate tests right now which is already being investigated.

Fix itself seems to have worked, as I see the workers on Cloud.

PR #2431, opened 2024-01-18
fix(fargate): convert to posix paths before sending files to s3

bernardobridge:
## Description

Paths to send to s3 are built based on a windows file system, but they are ran in ubuntu in the AWS containers. We need to account for that and convert the paths if needed.

## Pre-merge checklist


bernardobridge:
Windows test passing, confirming the fix: https://github.com/artilleryio/artillery/actions/runs/7573880065/job/20627137077

(was failing before)

PR #2439, opened 2024-01-24
fix(publish-metrics): otel tracing bug

InesNi:
# Description

Fix for OTel traces arriving inconsistently (in number and format) to observability platforms when running in Fargate 

## Context
Two different issues were observed, and usually at least one would happen:
1. All traces properly formed with the correct attributes and status, but some traces missing
2. Traces not properly formed - missing parent root spans, incorrect span status (errors not set on spans)
 
*Note: the issues would not happen on local runs, and issue 2. seems to only happen on Datadog.*

The issues seem to be a result of multiple factors:
- The wait for pending traces was not implemented in Playwright tracing possibly leading to triggering the `traceProvider ` shut down prematurely.
- In large tests with many traces the `maxExportBatchSize` and eventually the `maxQueueSize` could be reached in which case spans would be dropped

## Solution

This PR addresses the mentioned factors that seem to be causing the issues:

1. Waiting for pending traces is implemented in the Playwright tracing itself 
2. The defaults for `maxExportBatchSize`, `maxQueueSize` and the `scheduledDelayMillis` were increased, and the settings exposed to user for configuration if needed (this would allow users to adjust the number of traces that a batch can take before dropping them)
3. Additional 5sec wait is added prior to shutting down the `traceProvider` in order to wait for the exporters flush.

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/open-telemetry/tracing/base.js
+++ b/packages/artillery-plugin-publish-metrics/lib/open-telemetry/tracing/base.js
@@ -89,6 +89,9 @@ class OTelTraceBase {
     this.script = script;
     this.pendingRequestSpans = 0;
     this.pendingScenarioSpans = 0;
+    this.pendingPageSpans = 0;
+    this.pendingStepSpans = 0;
+    this.pendingPlaywrightScenarioSpans = 0;
   }
   setTracer(engine) {
     // Get and set the tracer by engine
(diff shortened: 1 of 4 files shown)


packages/artillery-plugin-publish-metrics/lib/open-telemetry/tracing/base.js
```diff
@@ -144,10 +148,21 @@ class OTelTraceBase {
 
   async cleanup() {
     while (this.pendingRequestSpans > 0 || this.pendingScenarioSpans > 0) {
```
hassy, line 150: Could we put a timeout on cleanup here? In case one of the conditions never evaluates to `true`. (This should really be done properly done at the runner level, but for now here is ok.)

hassy:
Is there a way to reduce the flush period of the Datadog agent?

Issue #2438, opened 2024-01-23
Playwright Feature suggestion: configure setTestIdAttribute


Georgegriff:
Playwright core exposes the function [getByTestId](https://playwright.dev/docs/locators#locate-by-test-id) this is a useful shorthand for ```page.locator(`[data-testid="${id}"]`)```

Not all apps however use `data-testid` as their test attribute, so playwright exposes the ability to configure this with [selectors.setTestIdAttribute](https://playwright.dev/docs/api/class-selectors#selectors-set-test-id-attribute).

Due to how `selectors.setTestIdAttribute.` works by setting up module scope variables it will only work if you import `import { selectors } from "playwright";` in the same package that launches the `chromium-browser`. In this case artillery-engine-playwright has playwright as a dependency so `selectors.setTestIdAttribute.` cannot be easily configured from the outside.

## Solution
This seems like a good case for a new config option in the `engines.playwright` options? something like `testIdAttribute`? and configure it in your engine code with an if statement for this config

I'm open to doing a PR for this if you would be interested.

## Workarounds:

Because we would rather not have to re-write our existing page object model code we have been using some workarounds.

It is technically possible to do this from the consumer code, but it feels like a hack to reach into the node modules of a dependency and causes issues with typescript.

```ts
import { selectors } from "artillery-engine-playwright/node_modules/playwright";
selectors.setTestIdAttribute("data-custom-id");
```

hassy:
A PR would be great @Georgegriff!

Georgegriff:
Thanks. I've open a PR https://github.com/artilleryio/artillery/pull/2441

PR #2441, opened 2024-01-24
feat(playwright): Added testIdAttribute configuration to playwright engine

Georgegriff:
As discussed in #2438 

Introduces the ability to configure the selector by exposing the same variables that playwright.config.ts usually exposes in use. Similar to defaultNavigationTimeout and defaultTimeout.

I would have added tests but I could find an existing setup for the playwright engine to contribute to.

Resolves: #2438


hassy:
Thanks @Georgegriff! We run Playwright examples in this repo as part of the test suite.

hassy:
AWS test failure is unrelated (due to #2289)

PR #2446, opened 2024-01-26
feat(fargate): set cloudwatch logs retention policy

bernardobridge:
## Description

Follow up of https://github.com/artilleryio/artillery/pull/2434

This sets retention policy for Cloudwatch logs. It sets an interval with a max number of retries to set the policy, so this isn't a blocking step and doesn't add additional time to the test unless something goes wrong.

Things that need review:
- Wasn't sure what was the ideal `LOGGROUP_RETENTION_DAYS` time should be.. Technically this is a change that could break workflows, if people are expecting a certain retention period. I put 6 months for now, but let's decide on the ideal time.
- I exposed the `process.env.ARTILLERY_LOGGROUP_RETENTION` environment variable so that this can be overridden. We may want this to be exposed in some other way, but for now it felt like a reasonable way to do it.

## Pre-merge checklist

- [ ] Does this require a changelog entry? Yes, and should highlight that it will update existing tests


packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
```diff
@@ -654,6 +658,15 @@ async function tryRunCluster(scriptPath, options, artilleryReporter) {
         await launchLeadTask(context);
       }
 
+      setCloudwatchRetentionOnInterval(
```
hassy, line 661: nitpick but whan does "OnInterval" mean here? if it refers to the implementation detail of using `setInterval` then imo it should be just `setCloudWatchRetention`

packages/artillery/lib/platform/aws-ecs/legacy/constants.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   SQS_QUEUES_NAME_PREFIX: 'artilleryio_test_metrics',
   S3_BUCKET_NAME_PREFIX: 'artilleryio-test-data',
   LOGGROUP_NAME: 'artilleryio-log-group',
+  LOGGROUP_RETENTION_DAYS: process.env.ARTILLERY_LOGGROUP_RETENTION || 180,
```
hassy, line 9:
```suggestion
  LOGGROUP_RETENTION_DAYS: process.env.ARTILLERY_LOGGROUP_RETENTION_DAYS || 180,
```

hassy: More descriptive name

hassy:
6 months sounds very reasonable. LGTM!

PR #2459, opened 2024-02-01
fix(fargate): start workers in parallel with lead task

hassy:
This reverts the behavior introduced in [1] where worker tasks may wait for up to 60s whilst the lead task is preparing npm dependencies.

The only benefit of waiting is to save up to 60s worth of Fargate compute costs for tests where the lead task fails to startup. This is rare.

Every test pays a penalty of having to wait for some time for the lead task to be ready before the rest of the workers even begin initializing however.

1. https://github.com/artilleryio/artillery/pull/2428


hassy:
Failing test is due to AWS Lambda zip size bug fixed in #2457

PR #2464, opened 2024-02-02
feat: add support for running on Fargate Spot

hassy:


bernardobridge:
Does this work for existing clusters? I thought we would have to leverage `checkTargetCluster` which is run before anyway, and then if no capacity provider is in the provided cluster, we could add it there.

If/When it does work for existing cluster, it might be worth just altering one of the tests to run on spot, to give us free testing of this feature?

hassy:
@bernardobridge existing clusters have to be recreated, will be in the docs

PR #2481, opened 2024-02-06
feat(fargate): use adot image for otel

InesNi:
# Description

Running AWS Distro for OpenTelemetry (ADOT) as a sidecar in Fargate in place of the Datadog agent in order to enable supporting multiple `publish-metrics` reporters rather than just Datadog.

## Context
Some observability platforms can't ingest OTLP data directly, but through either OpenTelemetry Collector or their own agent. Currently Artillery AWS Fargate runs support only one such platform - Datadog, by running the Datadog agent as a sidecar.

In order to enable providing Fargate support for other such reporters (primarily Cloudwatch reporter), in this PR the Datadog agent is replaced by ADOT. 

ADOT provides an array of [built-in components](https://github.com/aws-observability/aws-otel-collector?tab=readme-ov-file#adot-collector-built-in-components) including exporters for Datadog, AWS X-Ray, Cloudwatch, Splunk (SAML and SignalFX), Prometheus etc. Due to this implementing ADOT will enable providing tracing support for multiple platforms with one sidecar.

## Implementation

Multiple components needed to be implemented in order to configure and run ADOT:
- Access to a resolved config in the `run-cluster.js` file ( 3c03ac30acf49cd859d78d6c23cdce4e7fbe8b4e, 17d0268797504830d623864e3ce3b2a63fac7109, e2b9e68c02b23065692e9e24500957c046574ff3, ef54217ba8380bead065ac1b109661221338f588, 3c03ac30acf49cd859d78d6c23cdce4e7fbe8b4e) in order to:
  -  check if ADOT sidecar needs to be set, 
  - get the required reporter configurations 
- Vendor specific translator functions (7de3f851bc73e7a4edc4a44b27da2e26c43f881a): 
  - translate the reporter configuration into OTel Collector configuration (used by ADOT), 
  - provide any additional environment variables that need to be provided to ADOT 
- In order for ADOT to use our [custom configuration](https://aws-otel.github.io/docs/setup/ecs/config-through-ssm)(6aacc28a273cd0fbad00451c9c2f4ac4d5c2d06d):
  - SSM parameter whose value is the full custom configuration is created
  - The `AOT_CONFIG_CONTENT` variable that points to the created parameter is set in the ADOT task definition 
  - Any environment variables required by the custom config are provided in task overrides

  - The SSM parameter is deleted on cleanup

## Testing
- Manually tested with both `Playwright` and `HTTP` engine scenarios.
- Unit tests added on `publish-metrics` side for functions in `vendor-adot.js` 
- Unit test on CLI side added for  new `applyScriptChanges` function responsible for resolving templates in config
- e2e test added on CLI side


packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
```diff
@@ -978,19 +992,83 @@ async function createTestBundle(context) {
...
         } else {
+          context.fullyResolvedConfig = result.manifest.fullyResolvedConfig;
           return resolve(context);
         }
       }
     );
   });
 }
 
+async function createADOTDefinitionIfNeeded(context) {
```
InesNi, line 1010: As we agreed @hassy , I'm filtering the `publish-metrics` config in a separate function here before resolving the config and env vars for ADOT.

However maybe it can be moved into the `resolveADOTConfigSettings` to allow for better separation of concerns:  everything `publish-metrics` related would get done on `publish-metrics` side , and this function only does `run-cluster` type stuff (ssm, task definition, etc) ?  Let me know.

PR #2485, opened 2024-02-06
fix(fargate): propagate phase events to cli and cloud

bernardobridge:
## Description

Currently Fargate doesn't listen to these events and consume them. This PR makes the events available to the CLI command, deduplicates them [in a similar way to launch platform implementation](https://github.com/artilleryio/artillery/blob/main/packages/artillery/lib/launch-platform.js#L69-L70), and then emits the corresponding Artillery Cloud event so `phaseStarted`/`phaseCompleted` appears correctly in the Cloud.

### Testing

Has been tested against Artillery Cloud and phases appear correctly. Check screenshot:

Screenshot 2024-02-06 at 17 09 43

## Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/platform/aws-ecs/legacy/plugins/artillery-plugin-sqs-reporter/index.js
+++ b/packages/artillery/lib/platform/aws-ecs/legacy/plugins/artillery-plugin-sqs-reporter/index.js
@@ -75,6 +75,7 @@ function ArtillerySQSPlugin(script, events) {
     });
   });
 
+  //TODO: reconcile some of this code with how lambda does sqs reporting
   events.on('phaseStarted', (phaseContext) => {
     this.unsent++;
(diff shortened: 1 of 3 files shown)


packages/artillery/lib/platform/aws-ecs/legacy/plugins/artillery-plugin-sqs-reporter/index.js
```diff
@@ -75,6 +75,54 @@ function ArtillerySQSPlugin(script, events) {
     });
   });
 
+  events.on('phaseStarted', (phaseContext) => {
```
hassy, line 78: Is this code duplicated from somewhere else? if so would be useful to have a comment here to make refactoring it out later easier.

hassy:
lgtm

PR #2493, opened 2024-02-15
feat(playwright): use target as baseURL by default

bernardobridge:
## Description

`target` will now be used as `baseURL` by default. This change makes `artillery-engine-playwright` more consistent with Playwright (which allows the `baseURL` config option), and simplifies the job of transitioning from a Playwright codebase to an Artillery one, as you can simply set the `target` as the `baseURL` option from your Playwright tests, and keep the test scripts the same.

## Testing

- Tested existing test scripts from examples - they work as normal with the full URL, and work when we pass just the path (e.g. `await page.goto('/docs')`).

_Note: I'll change some of the examples to leverage this when a new CLI version contains the change._

## Pre-merge checklist

### DIFF
--- a/packages/artillery-engine-playwright/index.js
+++ b/packages/artillery-engine-playwright/index.js
@@ -76,7 +76,10 @@ class PlaywrightEngine {
         }
       }
 
-      const contextOptions = self.contextOptions || {};
+      const contextOptions = {
+        baseURL: self.target,
+        ...self.contextOptions
+      }
 
       let browser;
       if (self.useSeparateBrowserPerVU) {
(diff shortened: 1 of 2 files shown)


hassy:
lgtm!

alexandredepape:
I'm having trouble configuring this, this is my yaml:

```yaml
config:
  engines:
    playwright: {
      contextOptions: {
        baseURL: "http://localhost:3000"
      }
    }
  processor: "./main.ts"
scenarios:
  - engine: playwright
    testFunction: "main"

```
Because it says 

```bash
Error: No target specified and no environment chosen
```

What's the point of this PR if I have to repeat the target if it's already set in the baseURL ?

hassy:
@alexandredepape the point is that you need to set `target` which will be passed into Playwright config as the `baseURL`. You don't need to set it via `contextOptions`.

alexandredepape:
Got it, thanks 👌

PR #2514, opened 2024-02-20
fix(publish-metrics): route debug correctly for publish metrics reporters

InesNi:
# Description

When `DEBUG` is set to a vendor-specific reporter (from `publish-metrics`) if that reporter is using OTel in the background we want to log OTel reporters debug logs as well.

e.g. When user uses the Datadog reporter for tracing and sets `DEBUG=plugin:publish-metrics:datadog`, we want to log debug logs from OTel reporter as well as OTel is the one running the tracing.

## Implementation
- OpenTelemetry's internal diagnostic handler will run if `DEBUG` is set to any reporter, not just `open-telemetry`
- Artillery's `debug` will be dynamically set for metrics and traces, using the `type` property from config set to the reporters name.

In config translator functions `type` is set on both `traces` and `metrics` part of config so that classes responsible for metrics and traces can set the debug accordingly.  

## Testing
Tested manually, running multiple reporters and types of errors to trigger all debug logs and make sure the output is correct.

# Pre-merge checklist

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
+++ b/packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const debug = require('debug')('plugin:publish-metrics:open-telemetry');
 const vendorTranslators = require('./vendor-translators');
 const {
   diag,
@@ -47,11 +46,21 @@ class OTelReporter {
       // Setting traces to first traces configured
       if (translatedConfig.traces && !this.tracesConfig) {
         this.tracesConfig = translatedConfig.traces;
+
+        // Setting debug for traces
+        this.traceDebug = require('debug')(
+          `plugin:publish-metrics:${this.tracesConfig.type}`
+        );
       }
 
       // Setting metrics to first metrics configured
       if (translatedConfig.metrics && !this.metricsConfig) {
         this.metricsConfig = translatedConfig.metrics;
+
+        // Setting debug for metrics
+        this.metricDebug = require('debug')(
+          `plugin:publish-metrics:${this.metricsConfig.type}`
+        );
       }
       return translatedConfig;
     });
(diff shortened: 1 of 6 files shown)


packages/artillery-plugin-publish-metrics/lib/open-telemetry/vendor-translators.js
```diff
@@ -64,11 +64,14 @@ const vendorTranslators = {
     return otelTemplate(config, dynatraceTraceSettings);
   },
   'open-telemetry': (config) => {
-    let tracesConfig = config;
+    let newConfig = config;
```
InesNi, line 67: Changing name to `newConfig` as makes more sense, especially now when adding the `metrics` check as well.

packages/artillery-plugin-publish-metrics/lib/open-telemetry/index.js
```diff
@@ -24,7 +23,7 @@ context.setGlobalContextManager(contextManager);
 // DEBUGGING SETUP - setting the OpenTelemetry's internal diagnostic handler here to run when debug is enabled
 if (
   process.env.DEBUG &&
-  process.env.DEBUG === 'plugin:publish-metrics:open-telemetry'
+  process.env.DEBUG.includes('plugin:publish-metrics:')
```
InesNi, line 26: Setting OTels internal logger to run when `DEBUG` is set for any reporter, not just `open-telemetry`

packages/artillery-plugin-publish-metrics/lib/open-telemetry/tracing/base.js
```diff
@@ -75,20 +75,21 @@ class OTelTraceConfig {
...
-    debug('TracerProvider shutdown completed');
+    this.debug('TracerProvider shutdown completed');
   }
 }
 
 class OTelTraceBase {
   constructor(config, script) {
     this.config = config;
     this.script = script;
+    this.debug = require('debug')(`plugin:publish-metrics:${this.config.type}`);
```
InesNi, line 92: Setting `this.debug` here with reporter type will make it available to both HTTP and Playwright trace reporters.

packages/artillery-plugin-publish-metrics/lib/open-telemetry/metrics.js
```diff
@@ -45,6 +44,7 @@ class OTelMetricsReporter {
   }
 
   configure(config) {
+    this.debug = require('debug')(`plugin:publish-metrics:${this.config.type}`);
```
InesNi, line 47: Currently OTel reporter `metrics` doesn't support any vendor-specific reporters, but for future proofing we require dynamically.

bernardobridge (approved):

PR #2522, opened 2024-02-27
feat: async hooks and ESM

### LINKED ISSUE #1269: Support ES modules as processor
Currently there is some drive in parts of the JS community to switch from commonjs to ES modules.
One example is the http request library got, also used by artillery.

This change comes with a lot of incompatibilities / headaches and also artillery in the latest 2.x version (2.0.0-9) does not seem to support specifying ES modules as processor path.

Monkey-patching line `107` in `worker.js` (https://github.com/artilleryio/artillery/blob/master/lib/worker.js#L107)
from 
```
    let processor = require(processorPath);
```
to
```
    let processor = await import(processorPath);
```
seems to work for commonjs and ES modules.

As I do not feel 100% confident this might not have any unwanted side-effects I just want to mention this here as suggestion, without proof it is the right thing to do.

Would be great to see artillery in the future also support ES modules.

Looking forward to see version 2 finally land

hassy:
This PR add support for using async functions as hooks in HTTP and WebSocket engines, and the ability to load ES modules via `config.processor`.

**Hooks:**

- `beforeRequest`, `afterResponse`, `beforeScenario`, `afterScenario`, and `function` hooks in the HTTP engine
- `function` hook in the WebSocket engine

**ESM:**

- If the extension of a processor file is `.mjs`, it will be [`import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import)'ed rather than `require()`'d.

----

- [ ] Update to the docs needed
- [x] Changelog entry needed

----

Fixes #1269 #1384

### DIFF
--- a/packages/artillery/test/cli/run-typescript.test.js
+++ b/packages/artillery/test/cli/run-typescript.test.js
@@ -99,14 +99,13 @@ tap.test('Runs correctly when package is marked as external', async (t) => {
 tap.test(
   'Failure from a Typescript processor has a resolvable stack trace via source maps',
   async (t) => {
-    const [exitCode, output] = await execute([
+    const [_exitCode, output] = await execute([
       'run',
       '-o',
       `${reportFilePath}`,
       'test/scripts/scenarios-typescript/error.yml'
     ]);
 
-    t.equal(exitCode, 11, 'CLI should exit with code 11');
     t.ok(
       output.stdout.includes('error_from_ts_processor'),
       'Should have logged error from ts processor'
--- a/packages/artillery/test/cli/run-typescript.test.js
+++ b/packages/artillery/test/cli/run-typescript.test.js
@@ -95,33 +95,3 @@ tap.test('Runs correctly when package is marked as external', async (t) => {
 
   await deleteFile(bundleLocation);
 });
-
-tap.test(
-  'Failure from a Typescript processor has a resolvable stack trace via source maps',
-  async (t) => {
-    const [_exitCode, output] = await execute([
-      'run',
-      '-o',
-      `${reportFilePath}`,
-      'test/scripts/scenarios-typescript/error.yml'
-    ]);
-
-    t.ok(
-      output.stdout.includes('error_from_ts_processor'),
-      'Should have logged error from ts processor'
-    );
-
-    // // Search for the path
-    // const pathRegex = /\((.*?):\d+:\d+\)/;
-    // const match = output.stdout.match(pathRegex);
-
-    // // Extract the path if found
-    // const extractedPath = match ? match[1] : null;
-
-    // t.ok(
-    //   extractedPath.includes('.ts'),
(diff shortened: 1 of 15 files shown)


hassy:
Not sure what's going on with Socket Security here - none of the dependencies it listed as new are in fact new other than `detective-es6`.

packages/artillery/test/cli/run-typescript.test.js
```diff
@@ -99,14 +99,13 @@ tap.test('Runs correctly when package is marked as external', async (t) => {
...
   async (t) => {
-    const [exitCode, output] = await execute([
+    const [_exitCode, output] = await execute([
       'run',
       '-o',
       `${reportFilePath}`,
       'test/scripts/scenarios-typescript/error.yml'
     ]);
 
-    t.equal(exitCode, 11, 'CLI should exit with code 11');
```
bernardobridge, line 109: Why was the check for a failure exitCode removed? Maybe it shouldn't explicitly be 11, but it should still be non-zero, no?

packages/core/lib/runner.js
```diff
@@ -77,15 +77,25 @@ function loadEngines(
...
+
+    if (processorPath.endsWith('.mjs')) {
+      const exports = await import(processorPath);
+      script.config.processor = Object.assign(
+        {},
+        script.config.processor,
+        exports
+      );
+    } else {
+      // CJS (possibly transplied from TS)
```
bernardobridge, line 96: nit: `transpiled`

bernardobridge (dismissed):

bernardobridge:
Also is it possible to add a test for the `.mjs` extension working?

packages/artillery/test/cli/run-typescript.test.js
hassy: Exit code 11 was not explicitly set anywhere by Artillery itself either btw, it seemed to be set by Node.js when a worker crashes with an unhandled exception.

bernardobridge: Ah I see. Possibly this test should now check the number of vusers.failed then

bernardobridge:
@hassy 

FYI, was touching that part of the code to fix something else, so added `ESM` modules to the VS Code Extension schema already: https://github.com/artilleryio/artillery/pull/2525/commits/0e85c5b05a0e335e4d668cfe23bd471237ca5cc0

We can merge that when this is merged.

packages/artillery/test/cli/run-typescript.test.js
hassy: yep, that'd be a good test but does not belong in this specific test suite, as it's a TypeScript-specific test cases, not tests for the hooks API/behavior

bernardobridge: yeah I was wondering the same tbh

PR #2526, opened 2024-02-29
fix(cli): remove cpu/memory flag defaults to fix launch-config override

bernardobridge:
## Description

https://github.com/artilleryio/artillery/pull/2301 introduced a bug where `--launch-config` would no longer correctly set `memory` and `cpu`.

---

Funny enough, the `memory-hog` test (which was added later after that PR) was correctly catching this error, which is why it was failing - it had been wrongly modified to expect the non-zero code. #trustTheTests

I added some tests proving the fix and the feature work, leveraging the same `memory-hog` test.

## Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/cmds/run-fargate.js
+++ b/packages/artillery/lib/cmds/run-fargate.js
@@ -99,13 +99,11 @@ RunCommand.flags = {
   }),
   cpu: Flags.string({
     description:
-      'Set task vCPU on Fargate. Value may be set as a number of vCPUs between 1-16 (e.g. 4), or as number of vCPU units (e.g. 4096)',
-    default: '4'
+      'Set task vCPU on Fargate (defaults to 4 vCPU). Value may be set as a number of vCPUs between 1-16 (e.g. 4), or as number of vCPU units (e.g. 4096).'
   }),
   memory: Flags.string({
     description:
-      'Set task memory on Fargate. Value may be set as number of GB between 1-120 (e.g. 8), or as MiB (e.g. 8192)',
-    default: '8'
+      'Set task memory on Fargate (defaults 8 GB). Value may be set as number of GB between 1-120 (e.g. 8), or as MiB (e.g. 8192)'
   }),
   packages: Flags.string({
     description:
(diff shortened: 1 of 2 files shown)


packages/artillery/lib/cmds/run-fargate.js
```diff
@@ -99,13 +99,11 @@ RunCommand.flags = {
   }),
   cpu: Flags.string({
     description:
-      'Set task vCPU on Fargate. Value may be set as a number of vCPUs between 1-16 (e.g. 4), or as number of vCPU units (e.g. 4096)',
-    default: '4'
+      'Set task vCPU on Fargate (default 4vCPU). Value may be set as a number of vCPUs between 1-16 (e.g. 4), or as number of vCPU units (e.g. 4096).'
```
hassy, line 102:
```suggestion
      'Set task vCPU on Fargate (defaults to 4 vCPU). Value may be set as a number of vCPUs between 1-16 (e.g. 4), or as number of vCPU units (e.g. 4096).'
```

packages/artillery/lib/cmds/run-fargate.js
```diff
@@ -99,13 +99,11 @@ RunCommand.flags = {
...
   }),
   memory: Flags.string({
     description:
-      'Set task memory on Fargate. Value may be set as number of GB between 1-120 (e.g. 8), or as MiB (e.g. 8192)',
-    default: '8'
+      'Set task memory on Fargate (default 8GB). Value may be set as number of GB between 1-120 (e.g. 8), or as MiB (e.g. 8192)'
```
hassy, line 106:
```suggestion
      'Set task memory on Fargate (defaults 8 GB). Value may be set as number of GB between 1-120 (e.g. 8), or as MiB (e.g. 8192)'
```

packages/artillery/lib/cmds/run-fargate.js
hassy: Space between value & unit.

"(default 4 vCPU)" looks odd, went for "(defaults to 4 vCPU)" vs "(default: 4 vCPU)" here

packages/artillery/lib/cmds/run-fargate.js
```diff
@@ -99,13 +99,11 @@ RunCommand.flags = {
   }),
   cpu: Flags.string({
     description:
-      'Set task vCPU on Fargate. Value may be set as a number of vCPUs between 1-16 (e.g. 4), or as number of vCPU units (e.g. 4096)',
-    default: '4'
+      'Set task vCPU on Fargate (defaults to 4vCPU). Value may be set as a number of vCPUs between 1-16 (e.g. 4), or as number of vCPU units (e.g. 4096).'
```
hassy, line 102: spaces still missing 😆

PR #2527, opened 2024-02-29
ci(worker): fix broken ecs worker image version push

bernardobridge:
## Description

I just noticed that all image versions generated by AWS tests seem to be pointing to the `main` branch ref 💀 (https://github.com/artilleryio/artillery/actions/workflows/run-aws-tests.yml). I actually believe it is currently overriding the latest image in main. So the tests run against the right version of the code (which is why they pass), but the main branch image version gets overridden each time.

Important to fix ASAP.

P.S. This hasn't affected any `production` CLI release versions - fortunately there were pushes between each release and the next commit that had `run-aws-tests` each time.

## Pre-merge checklist

### DIFF
--- a/.github/workflows/docker-ecs-worker-image.yml
+++ b/.github/workflows/docker-ecs-worker-image.yml
@@ -19,6 +19,9 @@ permissions:
 jobs:
   build_docker_image:
     runs-on: ubuntu-latest
+    env:
+      # From PR's this is overridden by a head ref by the caller workflow. defaults to commit sha when not passed (e.g. workflow_dispatch or call from main branch)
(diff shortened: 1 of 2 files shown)


bernardobridge:
Note: https://github.com/artilleryio/artillery/pull/2527/commits/35c3ac437585d9ea51d0bb7031cf61aa2e91f6fe increases test timeout temporarily due to failures observed in CI from cancellation (e.g. https://github.com/artilleryio/artillery/actions/runs/8099617233/job/22135921668). These tests will be parallelised in PR's very soon, but in the meantime best to be generous with their runtime to prevent chasing CI failures that aren't relevant.

PR #2563, opened 2024-03-12
test(dep): update tap test runner to latest (18.7.1)

bernardobridge:
## Description

We have a few dependency updates to do with our dev dependencies. This one was the biggest one, as there are some changes between tap 16 and 18. 

Ended up making a couple of quality of life improvements too:
- Set disable coverage and color options in a package.json tap config, so we don't have to add `--color` and coverage flags on every time
- Also set `export ARTILLERY_TELEMETRY_DEFAULTS='{\"source\":\"test-suite\"}'` in the tap config

As a result, the npm scripts are now a lot shorter and it's harder to forget default flags/variables 👌 

## Pre-merge checklist


bernardobridge:
Closed in favour of https://github.com/artilleryio/artillery/pull/2797

PR #2565, opened 2024-03-12
fix: emit stats event for all periods for which there's data

hassy:
Depending on the exact time a test completes (all workers finish) there may be more than one time period that hasn't been reported yet.

This PR makes sure that when we exit (flushAll is set), we emit reports for all outstanding periods.

### DIFF
--- a/packages/artillery/lib/launch-platform.js
+++ b/packages/artillery/lib/launch-platform.js
@@ -321,9 +321,11 @@ class Launcher {
     }
 
     // We always look at the earliest period available so that reports come in chronological order
-    const earliestPeriodAvailable = Object.keys(this.metricsByPeriod)
+    const unreportedPeriods = Object.keys(this.metricsByPeriod)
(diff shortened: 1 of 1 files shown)


packages/artillery/lib/launch-platform.js
```diff
@@ -356,58 +358,49 @@ class Launcher {
...
+      (allWorkersReportedForPeriod || waitedLongEnough)
     ) {
+      this.emitIntermediatesForPeriod(earliestPeriodAvailable);
       // TODO: autoscaling. Handle workers that drop off or join, and update count
+    } else {
+      debug('Waiting for more workerStats before emitting stats event');
+    }
+  }
 
-      if (flushAll) {
```
hassy, line 365: This is now in the `emitIntermediatesForPeriod()` function.

hassy:
Failing test is ADOT-specific and unrelated to the changes here (also failed in #2555). Will look at separately.

PR #2603, opened 2024-03-27
feat: initial Playwright tracing support

hassy:
This PR adds the following:

* Capture Playwright traces for failing VUs and store them on the filesystem (locally or in Fargate workers)
* Upload them to Artillery Cloud (an account with an API key is required)

This is early experimental support for it and not something ready for production use yet.


hassy:
Socket Security reporting is broken again.

@socketsecurity ignore-all

PR #2607, opened 2024-03-28
feat(publish-metrics): implement replaceSpanNameRegex setting

InesNi:
# Description

- Based on discussion [here](https://github.com/artilleryio/artillery/discussions/2592).

- Implementing `replaceSpanNameRegex` config option in tracing for all reporters that will allow replacing patterns in span names by specifying the regex pattern and the replacements string.

- This is mostly provided for the purpose of quantization, allowing users to group spans that contain randomised patterns but are, for the purpose of analysis, essentially the same.

## Configuration

```yaml
  plugins:
    publish-metrics:
      - type: honeycomb
        apiKey: "{{ $env.HC_API_KEY }}"
        dataset: "my-service"
        replaceSpanNameRegex:
          - pattern: get_user_[a-zA-Z0-9]+
            as: get_user_id
```

- In the example above page span in Playwright tracing `Page: https://get_user_jd83rh47fhfaAjh7snjqh` would be changed to `Page: https://get_user_id`.

## Implementation
- `replaceSpanNameRegex` method is added to the `OTelTraceBase` class making it available for both HTTP and Playwright engine tracing.
- Upon each span creation if `replaceSpanNameRegex` is set in config, the span name will be modified if pattern is found.
- The schema for every reporter using OTel tracing has been updated

## Testing
- A unit test has been added for the `replaceSpanNameRegex`
- Manually tested with both engines and spans on all levels

# Pre-merge checklist

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/open-telemetry/tracing/base.js
+++ b/packages/artillery-plugin-publish-metrics/lib/open-telemetry/tracing/base.js
@@ -106,18 +106,23 @@ class OTelTraceBase {
   // Sets the tracer by engine type, starts the scenario span and adds it to the VU context
   startScenarioSpan(engine) {
     return function (userContext, ee, next) {
-      const span = this[`${engine}Tracer`].startSpan(
-        userContext.scenario?.name || `artillery-${engine}-scenario`,
-        {
-          startTime: Date.now(),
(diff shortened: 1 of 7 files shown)


joshuscurtis:
@InesNi Thanks for adding this feature - very much appreciated! 👍

PR #2635, opened 2024-04-11
feat(examples): add an example for pulling data from redis

bernardobridge:
## Description

Adds example for popular question regarding using unique test data per VU.


hassy:
lgtm

Issue #2661, opened 2024-04-22
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME] when using a artillery runner file with a `.mjs` extension


thrandale:
Artillery is supposed to support ES Modules. Per the documentation, the way of doing this is to have your `artilleryRunner` file have the `.mjs` extension instead of the `.js` extension. I am using an `artillery.yaml` file to pass in the processor:
``` yaml
config:
  ...
  processor: "artilleryRunner.mjs"
  ...
```
When running artillery it gives the error `Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'`

In doing some tracing, this error is Coming from `core/lib/runner.js` when it is trying to load the `artilleryRunner.mjs` file. Per [this](https://stackoverflow.com/questions/69665780/error-err-unsupported-esm-url-scheme-only-file-and-data-urls-are-supported-by#:~:text=That%20actually%20is,g.%2C%20C%3A)) Stack Overflow Post: `the import() function [only works with file:// and data:// URLs](https://nodejs.org/api/errors.html#err_unsupported_esm_url_scheme), but path.resolve() only returns an absolute path (not a URL)`

This can be fixed by doing a `pathToFileURL()` before doing the `import()` of the module

Version info: 
```
Artillery: 2.0.10
Node.js:   v18.17.1
OS:        win32
```

ShaunLawrie:
This is impacting me too. Windows paths are getting detected as urls

ggecy:
This is making the artillery completely unusable on windows if we have a typescript configured to compile playwright tests to esm modules.

cwongmt:
tests.yml

```
config:
  target: 'https://something.com'
  phases:
    - duration: 1
      arrivalRate: 1
  processor: '../auth.mjs'
  variables:
    user: 'tester'
scenarios:
  - name: 'Get'
    flow:
      - get:
          beforeRequest: 'auth'
          url: '/path'
          headers:
            Authorization: 'Bearer {{ token }}'
```
auth.mjs
```
export async function auth(requestParams, context, events) {
  const user = context.vars.user;
  console.log(`user: ${user}`);
  context.vars.token = 'test';
}
```

Facing the same error with above settings. This error renders it completely unusable on Windows.

bernardobridge:
[...]

PR #2662, opened 2024-04-22
fix(core): import mjs path correctly when running on windows

thrandale:
This fixes issue #2661

### DIFF
--- a/packages/core/lib/runner.js
+++ b/packages/core/lib/runner.js
@@ -18,6 +18,7 @@ const isIdlePhase = require('./is-idle-phase');
 const createReader = require('./readers');
 const engineUtil = require('@artilleryio/int-commons').engine_util;
 const wl = require('./weighted-pick');
+const { pathToFileURL } = require('url');
 
 const Engines = {
   http: require('./engine_http'),
@@ -86,7 +87,8 @@ async function loadProcessor(script, options) {
     );
 
     if (processorPath.endsWith('.mjs')) {
-      const exports = await import(processorPath);
+      const fileUrl = pathToFileURL(processorPath);
+      const exports = await import(fileUrl.href);
       script.config.processor = Object.assign(
         {},
         script.config.processor,


hassy:
thanks @thrandale! Is the issue due to a missing relative path? ie. in your example `./artilleryRunner.mjs` should work as expected (if the MJS file is in the same directory as the YAML script). There's a test for loading ESM here, would you mind updating that please? https://github.com/artilleryio/artillery/blob/242a7964934851bf77e53eebf44e6312de723663/packages/artillery/test/cli/async-hooks-esm.test.js#L10

thrandale:
@hassy Even using a relative path, it still fails with the same error. It is odd that it is not failing in the CI. I believe this is only an issue on Windows. Does the CI run the windows tests?

ggecy:
> @hassy Even using a relative path, it still fails with the same error. When it does the `path.resolve()`, that resolve either an absolute or relative path to an absolute path. It is odd that it is not failing in the CI. I believe this is only an issue on Windows. Does the CI run the windows tests?

I can confirm this - we have this problem only on windows, I can run on macos without error.

bernardobridge:
Hi @thrandale 👋 ! Thanks for the contribution. We have now added windows tests to CI, so we could confirm fixes like this.

I've opened https://github.com/artilleryio/artillery/pull/3293 with your fix and ran our full suite against it, and it seems to work 👍 . If you want to rebase this branch against `main`, I can close my PR and merge this one once the test suite runs here too. If you do that in the next couple of days (as we aim to release on Monday), I'll use your PR. Otherwise I'll use my PR and reference you in the release anyway.

Thanks again!

thrandale:
Thanks @bernardobridge. I have rebased my branch off `main`.

bernardobridge:
Hi @thrandale, thanks for that. I'm merging this as the tests as passing (they appear red, but it's a bug that happens when running against contributor PR's).

Thanks again for the fix! It'll be out in 2.0.19 next week.

Issue #1695, opened 2022-11-22
Warning: multiple batches of metrics for period


juanig1:
Version info: 
```
latest
```

Running this command: 
```
artillery run   --platform aws:lambda   --platform-opt region=us-east-1 --count 1   ~/scripts/hello/hello.yaml
```
also happens with [this](https://github.com/artilleryio/artillery/issues/1617) setup if the server is SLOW (like 20-30s slow)

I expected to see this happen:

Standard output with not warnings (1 request only)

Instead, this happened:

Execution went ok but got `Warning: multiple batches of metrics for period XXXX`
Does not seem to happen with a single worker, and also summary is not missing metrics.
Guess: I assume we are reporting empty from a worker that has no arrivals configured?
Files being used:
`hello.yaml`
```
config:
  target: "http://0.0.0.0:8080"
  phases:
    - duration: 1
      arrivalRate: 1
scenarios:
  - name: "get hello"
    flow:
      - get:
          url: "/"
          expect:
            - statusCode: 200
```

See https://github.com/artilleryio/artillery/discussions/1587 for original discussion

Zacaria:
hello !
I'm also interested in understand this issue

madhugod:
Had this issue with a `beforeRequest` hook function that wasn't `async`

I added the `async` keyword and for some reason it solved the issue

stvnwrgs:
It seems like this behavior leads to the cluster not shutting down anymore when running with fargate and playwright as engine. This makes it unusable for any CI kind of execution due to the high potential of financial risk.

```yaml
- duration: 60s
   arrivalRate: 5
   maxVusers: 5
```

```
Waiting for workers to start: pending: 5 [14:09:39]
Waiting for workers to start: pending: 5 [14:09:49]
Waiting for workers to start: pending: 5 [14:09:59]
Waiting for workers to start: pending: 3 / ready: 2 [14:10:09]
Waiting for workers to come online... [14:10:09]
Workers are running, waiting for reports... [14:11:11]
Warning: multiple batches of metrics for period 1713874270000 2024-04-23T12:11:10.000Z
Warning: multiple batches of metrics for period 1713874340000 2024-04-23T12:12:20.000Z
Warning: multiple batches of metrics for period 1713874370000 2024-04-23T12:12:50.000Z
Warning: multiple batches of metrics for period 1713874580000 2024-04-23T12:16:20.000Z
⠙ Stopping test run (SIGINT received)...
```

PR #2665, opened 2024-04-23
fix: (1695) endless runs when metrics are not provided correctly

stvnwrgs:
## Summary

We had several issues with running our tests locally and with fargate using playwright as engine. It seems, that bad code of the tests and things we could not find, lead to the behavior that our workers never ended. This only ocoured when  `Warning: multiple batches of metrics for period XXXX` happened.

https://github.com/artilleryio/artillery/issues/1695 describes the problem.

I would like to not only test this locally, but also with fargate but could not find any documentation on how to use custom containers in fargate. Any hints?

## Changes

This pull request primarily focuses on improving the handling of metrics reporting in the `Launcher` and `SqsReporter` classes in the `launch-platform.js` and `sqs-reporter.js` files respectively. The changes include the addition of a fallback mechanism for unreported periods, better management of the earliest unreported period, and improved condition checks for normal flush case.

Improvements to metrics reporting:

* [`packages/artillery/lib/launch-platform.js`](diffhunk://#diff-ba4e2ce001b17c74a286fa159f37fbd309192740a60f3448e95b5da6aaf4cd8fR324-R340): In the `Launcher` class, the handling of unreported periods was improved by adding a fallback mechanism when the earliest unreported period is undefined. 

* [`packages/artillery/lib/launch-platform.js`](diffhunk://#diff-ba4e2ce001b17c74a286fa159f37fbd309192740a60f3448e95b5da6aaf4cd8fR351-R386): The management of the earliest unreported period was further improved in the `Launcher` class. The variable `allWorkersReportedForPeriod` was moved up in the code and now checks if the length of `metricsByPeriod` for the `earliestUnreportedPeriod` equals the `count`. The `waitedLongEnough` variable was also moved up and now checks if the current time minus the `availableEarliestPeriod` is greater than `MAX_WAIT_FOR_PERIOD_MS`. These changes make the code more readable and maintainable.

* [`packages/artillery/lib/platform/aws-ecs/legacy/sqs-reporter.js`](diffhunk://#diff-6beaa8f46fbc7470ce03ff2092283ac13b1754c0cc8fc450f908a851f562f284L71-R81): Similar changes were made in the `SqsReporter` class. The handling of unreported periods was improved by adding a fallback mechanism when the earliest unreported period is undefined. If no earliest unreported period is found, a debug message is logged and the function returns. This change ensures that the function does not proceed without valid data, preventing potential errors.

* [`packages/artillery/lib/platform/aws-ecs/legacy/sqs-reporter.js`](diffhunk://#diff-6beaa8f46fbc7470ce03ff2092283ac13b1754c0cc8fc450f908a851f562f284L92-R130): The condition checks for the normal flush case were improved in the `SqsReporter` class. This change simplifies the condition checks and makes the code more readable.fix: NaN error in waitedLongEnough

### DIFF
--- a/packages/artillery/lib/launch-platform.js
+++ b/packages/artillery/lib/launch-platform.js
@@ -315,20 +315,24 @@ class Launcher {
   }
 
   async flushIntermediateMetrics(flushAll = false) {
-    if (Object.keys(this.metricsByPeriod).length === 0) {
-      debug('No metrics received yet');
-      return;
-    }
-
     // We always look at the earliest period available so that reports come in chronological order
+    // filter over individual periods, find not reported ones and sort them
     const unreportedPeriods = Object.keys(this.metricsByPeriod)
       .filter((x) => this.periodsReportedFor.indexOf(x) === -1)
       .sort();
 
-    const earliestPeriodAvailable = unreportedPeriods[0];
+    // fallback if undefined
+    const earliestUnreportedPeriod = unreportedPeriods[0] || 0;
+
+    if (!earliestUnreportedPeriod) {
+      debug('No metrics received yet');
+      return;
+    }
 
     // TODO: better name. One above is earliestNotAlreadyReported
     const earliest = Object.keys(this.metricsByPeriod).sort()[0];
+
+    // double check with unreportedPeriods?
     if (this.periodsReportedFor.indexOf(earliest) > -1) {
       global.artillery.log(
         'Warning: multiple batches of metrics for period',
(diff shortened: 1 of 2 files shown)


hassy:
Sounds good @stvnwrgs, thank you for taking the time! I'll review the changes later today.

The worker image used on Fargate can be overridden via [`WORKER_IMAGE_URL` env var](https://github.com/artilleryio/artillery/blob/main/packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js#L1145), but it requires that you build and push a new worker image to an accessible Docker registry somewhere. We have a GitHub Actions workflow that makes that easier but it has to be approved by a committer first.

stvnwrgs:
@hassy Awesome! I will try that and post if something wont work correctly. 

Btw: I focused on the `Launcher` but did the `SQS` stuff too but without testing it.

stvnwrgs:
@hassy The changes do not work for Fargate. The jobs are started, but getting closed after a short time again without providing the metrics. 

Is there any documentation about the general behavior how the communication between the cli, workers, queue and reporter works?

hassy:
I'll take a look @stvnwrgs!
 
> Is there any documentation about the general behavior how the communication between the cli, workers, queue and reporter works?

Not at the moment. At the highest level:

- CLI creates an SQS queue and passes its URL to workers on startup
- Workers send messages to that SQS queue for each `stats` event (and a couple of others)
- The CLI receives and aggregates those messages (in sqs-reporter)
- It then emits another set of `stats` messages - these are the ones that get picked up by Artillery plugins
- Console reporter is an Artillery plugin, so it subscribes to those events and converts those stats objects to text output

stvnwrgs:
Closing, as solution did not work on fargate and we moved to k6.

hassy:
Thanks for the update @stvnwrgs! I'm still going to try to pick up the changes in the PR as they do sound very useful. Just haven't been able to grab some time for a thorough review last week. 

Did you move to using k6 Browser instead of Playwright?

PR #2666, opened 2024-04-23
feat(cli): add deprecation notice to report cmd

InesNi:
# Description

Adding deprecation notice to `report` command output.

Screenshot 2024-04-23 at 16 57 40

### DIFF
--- a/packages/artillery/lib/cmds/report.js
+++ b/packages/artillery/lib/cmds/report.js
@@ -7,6 +7,7 @@ const telemetry = require('../telemetry').init();
 
 class ReportCommand extends Command {
   async run() {
+    console.log(deprecationNotice);
     telemetry.capture('report generate');
     const { flags, args } = await this.parse(ReportCommand);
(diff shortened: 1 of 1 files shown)


packages/artillery/lib/cmds/report.js
```diff
@@ -35,6 +36,21 @@ ReportCommand.flags = {
   })
 };
 
+const deprecationNotice = `
+\x1b[34m┌───────────────────────────────────────────────────────────────────────┐
```
hassy, line 40: We should use `chalk` here (already used by the CLI) which e.g. takes care of checking if output is being redirected in which case ANSI codes are not included in the output.

packages/artillery/lib/cmds/report.js
```diff
@@ -7,6 +7,7 @@ const telemetry = require('../telemetry').init();
 
 class ReportCommand extends Command {
   async run() {
+    console.log(deprecationNotice);
```
hassy, line 10: This should probably be `console.error()` instead

PR #2670, opened 2024-04-24
Change artillery probe to skytrace probe in README.md

theodorekouzelis:
The command "artillery probe" doesn't seem to exist, but "skytrace probe" does and produces the pictured output.


hassy:
thanks @theodorekouzelis!

PR #2674, opened 2024-04-25
feat: use container images in lambda

bernardobridge:
## Description

TBA - opening this now in case early feedback is wanted. Reviewing commit by commit is recommended.

To-do:

- [x] Refactor Lambda to make it easier to keep backwards compatibility (https://github.com/artilleryio/artillery/commit/c89448627eeae6949250257e05b2993940b94408, https://github.com/artilleryio/artillery/commit/2b39f09c66e42bad72f1c90e84f9a541213ed30f)
- [x] Implement Lambda using container images opted-in (https://github.com/artilleryio/artillery/commit/66498cfd473714cb0b66a5a7a4e1e2bdd0b9fef1)
- [x] Reuse same Docker image for Lambda and Fargate (https://github.com/artilleryio/artillery/commit/43d954a931e004b5bd6db2d1236e93b80ddb3e94, https://github.com/artilleryio/artillery/commit/f860db5ee2b6aa6b065fd9da55efbbdf0bc0af3f, https://github.com/artilleryio/artillery/commit/04015c33484f1d69cbad8231fc19f42512afe106)
- [ ] Reuse same Lambda function (WIP - need to figure out mechanism to determine version)
- [ ] Deprecation notice for regular Lambda support
- [ ] Improve console messages when packaging

### DIFF
--- a/packages/artillery/lib/cmds/run-lambda.js
+++ b/packages/artillery/lib/cmds/run-lambda.js
@@ -34,6 +34,10 @@ class RunLambdaCommand extends Command {
       flags['platform-opt'].push(`subnet-ids=${flags['subnet-ids']}`);
     }
 
+    if (flags.container) {
+      flags['platform-opt'].push('container=true');
+    }
+
     flags.platform = 'aws:lambda';
 
     RunCommand.runCommandImplementation(flags, argv, args);
@@ -59,6 +63,10 @@ RunLambdaCommand.flags = {
   count: Flags.string({
     // locally defaults to number of CPUs with mode = distribute
     default: '1'
+  }),
+  container: Flags.boolean({
+    description: 'Run the test in a container (recommended)',
+    default: false
   }),
   architecture: Flags.string({
     description: 'Architecture of the Lambda function',
--- a/packages/artillery/lib/platform/aws-ecs/worker/Dockerfile
+++ b/packages/artillery/lib/platform/aws-ecs/worker/Dockerfile
@@ -32,4 +32,4 @@ COPY ./packages/artillery/lib/platform/aws-ecs/worker/helpers.sh /artillery/help
 
 RUN chmod +x /artillery/loadgen-worker
 
-ENTRYPOINT ["/artillery/loadgen-worker"]
+ENTRYPOINT ["/artillery/packages/artillery/bin/run"]
--- a/packages/artillery/lib/platform/aws-ecs/worker/Dockerfile
+++ b/packages/artillery/lib/platform/aws-ecs/worker/Dockerfile
@@ -2,34 +2,57 @@
 # NOTE: Version we use here needs to be kept consistent with that in
 # artillery-engine-playwright.
 # ********************************
-FROM mcr.microsoft.com/playwright:v1.43.1
-LABEL maintainer="team@artillery.io"
+FROM mcr.microsoft.com/playwright:v1.43.1 as build-image
+
+# https://steele.blue/playwright-on-lambda/
+# https://blog.carlosnunez.me/post/scraping-chromium-lambda-nodeless-zerostress/
 
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && \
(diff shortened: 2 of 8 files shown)


packages/artillery/lib/cmds/run-lambda.js
```diff
@@ -60,6 +64,10 @@ RunLambdaCommand.flags = {
     // locally defaults to number of CPUs with mode = distribute
     default: '1'
   }),
+  container: Flags.boolean({
+    description: 'Run the test in a container (recommended)',
```
hassy, line 68:
```suggestion
    description: 'Run the test in a container (experimental)',
```

packages/artillery/lib/platform/aws-ecs/worker/Dockerfile
```diff
@@ -2,34 +2,57 @@
 # NOTE: Version we use here needs to be kept consistent with that in
 # artillery-engine-playwright.
 # ********************************
-FROM mcr.microsoft.com/playwright:v1.43.1
-LABEL maintainer="team@artillery.io"
+FROM mcr.microsoft.com/playwright:v1.43.1 as build-image
+
+# https://steele.blue/playwright-on-lambda/
```
hassy, line 7: What's the relevance of these URLs? Can we add relevant comments inline if appropriate instead?

packages/artillery/lib/cmds/run-lambda.js
hassy:
```suggestion
    description: 'Use a container image for Lambda (experimental)',
```

Think we need to be more specific here than "run the test in a container". Lambda users will probably be aware of the two different packaging methods.

packages/artillery/lib/platform/aws-ecs/worker/Dockerfile
```diff
@@ -2,34 +2,57 @@
 # NOTE: Version we use here needs to be kept consistent with that in
 # artillery-engine-playwright.
 # ********************************
-FROM mcr.microsoft.com/playwright:v1.43.1
-LABEL maintainer="team@artillery.io"
+FROM mcr.microsoft.com/playwright:v1.43.1 as build-image
```
hassy, line 5:
```suggestion
FROM mcr.microsoft.com/playwright:v1.43.1
```

Doesn't seem to be a multi-stage (yet?)

packages/artillery/lib/platform/aws-lambda/dependencies.js
```diff
@@ -197,15 +208,118 @@ const createAndUploadLambdaZip = async (
...
+      .promise();
+
+    debug(`Uploaded ${key}`);
+    return;
+  } catch (err) {
+    throw err;
+  }
+}
+
+async function _syncS3(bomManifest, testRunId, bucketName) {
```
hassy, line 253: This (I think) would be the exact same logic as what we already have for Fargate. Would be better to use that code instead (refactoring it to make it reusable if needed) rather than duplicating some of it here?

bernardobridge: Yes that's the plan, but doing that right now risks potentially breaking Fargate and Lambda. To minimise the risk of the change, I'd rather refactor and unify that logic separately.

packages/artillery/lib/platform/aws-lambda/index.js
```diff
@@ -744,6 +628,60 @@ class PlatformLambda {
...
+
+    try {
+      const res = await lambda
+        .getFunctionConfiguration({
+          FunctionName: functionName
+        })
+        .promise();
+
+      if (res.MemorySize != memorySize) {
+        console.log(
```
hassy, line 670:
```suggestion
        debug(
```

Should be a debug log IMO - since function reuse is supposed to be transparent to the user.

packages/artillery/lib/platform/aws-ecs/worker/Dockerfile
bernardobridge: these were more for getting Playwright to work with Lambda. I'll keep them off for now, since I will address that separately in the end, and add it back if necessary

PR #2703, opened 2024-05-06
feat: remove is-builtin-module

43081j:
The node versions we support as per our engines constraint already have a built in function for determining if a module is built in.

This removes the dependency and uses the built in function instead.

## Pre-merge checklist


hassy:
thanks @43081j!

Issue #2673, opened 2024-04-25
Artillery ignores parenthesis in scenarios names


HugoImaios:
Hello Artillery team !
Just a little issue with scenario names : it seems like they simply ignore parenthesis.

Version info:

```
Artillery: 2.0.10
```

Running this command:

```
yarn artillery run config.yaml --scenario-name="Get 2 animal pictures (but with parenthesis)"
```

I expected to see this happen:

Artillery running my script.

Instead, this happened:

`worker error, id: 1 Error: Scenario Get 2 animal pictures (but with parenthesis) not found in script. Make sure your chosen scenario matches the one in your script exactly.`

Files being used:

`config.yaml`

```
config:
    # This is a test server run by team Artillery
    # It's designed to be highly scalable
    target: "http://asciiart.artillery.io:8080"
    phases:
        - duration: 60
          arrivalRate: 1
          rampTo: 5
          name: Warm up phase
scenarios:
  - name: Get 1 animal pictures
    flow:
        - loop:
            - get:
                url: "/dino"
          count: 100
  - name: Get 2 animal pictures (but with parenthesis)
    flow:
        - loop:
            - get:
                url: "/dino"
            - get:
                url: "/pony"
          count: 100
```

HugoImaios:
More informations :

If the two scenarios have the same name *except the parenthesis*, for example:

```
scenarios:
  - name: Get 1 animal pictures
    flow:
        - loop:
            - get:
                url: "/dino"
          count: 100
  - name: Get 1 animal pictures (but with parenthesis)
    flow:
        - loop:
            - get:
                url: "/dino"

          count: 100
```

and we run `yarn artillery run config.yaml --scenario-name="Get 1 animal pictures"`, we get another error :

```
Multiple scenarios for Get 1 animal pictures found in script. Make sure you give unique names to your scenarios in your script.
```
which leads to think the parenthesis are stripped off the scenario name

HugoImaios:
I also tried with quoted name in yaml, with no more success :

```
  - name: "Get 3 animal pictures (but with parenthesis, and quotes)"
    flow:
        - loop:
            - get:
                url: "/dino"
            - get:
                url: "/pony"
            - get:
                url: "/armadillo"
          count: 100
```

`artillery run config.yaml --scenario-name="Get 3 animal pictures (but with parenthesis, and quotes)"` gives :

```
[...]

PR #2709, opened 2024-05-07
fix(core): allow scenario name exact match without using regex

bernardobridge:
## Description

Addresses https://github.com/artilleryio/artillery/issues/2673 by allowing regex or exact match. I think this allows the maximum flexibility for the flag without breaking any existing workflows.

## Pre-merge checklist

- [ ] Does this require an update to the docs? Yes (need to explain how the feature works - regex vs exact match)


HugoImaios:
seems like the more pragmatic choice indeed ! Thanks @bernardobridge !

PR #2750, opened 2024-05-15
fix(lambda): make Lambda containers work with hierarchical deps

bernardobridge:
## Description

This copies over a Fargate test, then fixes the issue it uncovered. There's a few small things it solves to improve the experience overall:

- https://github.com/artilleryio/artillery/pull/2750/commits/50ba526ed0bd47027762e9ef21c6fdce273ec13e : Adds the test
- https://github.com/artilleryio/artillery/pull/2750/commits/86d2efae3c75569d31597f7aea431bd24d1e8356 : Adds dependency tree logic from Fargate
- https://github.com/artilleryio/artillery/pull/2750/commits/17a20593dc294b32b8bdb5e59753e02c11100882 : Appropriately catch errors from S3 sync - currently it would continue without any user feedback
- https://github.com/artilleryio/artillery/pull/2750/commits/4e70e1e0cb8ea99590beccc05d8d20455e65892c : look for files recursively when doing the file count check (files uploaded == files available on worker)
- https://github.com/artilleryio/artillery/pull/2750/commits/6823644345e370f843caaa4e65b6fcba27b53243 : start sending message and stack explicitly in errors. When sending the full error object a JSON.stringify is done which results in the error object being empty. This way the user now has clear feedback as originally intended.

## Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/create-bom/create-bom.js
+++ b/packages/artillery/lib/create-bom/create-bom.js
@@ -10,6 +10,7 @@ const A = require('async');
 const debug = require('debug')('bom');
 const _ = require('lodash');
 const Table = require('cli-table3');
+const { getCustomJsDependencies } = require('../platform/aws-ecs/legacy/bom');
 
 const { readScript, parseScript } = require('../util');
 
@@ -163,25 +164,6 @@ function getCustomEngines(context, next) {
   context.npmModules = context.npmModules.concat(enginePackages);
 
   return next(null, context);
-}
-
-function getCustomJsDependencies(context, next) {
-  if (
-    context.opts.scriptData.config &&
-    context.opts.scriptData.config.processor
-  ) {
-    const procPath = path.resolve(
-      path.dirname(context.opts.absoluteScriptPath),
-      context.opts.scriptData.config.processor
-    );
-    context.localFilePaths.push(procPath);
-
-    debug('got custom JS dependencies');
-    return next(null, context);
-  } else {
-    debug('no custom JS dependencies');
-    return next(null, context);
-  }
 }
 
 function getVariableDataFiles(context, next) {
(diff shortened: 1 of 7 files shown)


hassy:
For this - https://github.com/artilleryio/artillery/commit/881507f826136c2dda417bc82139b1843b8f9ec3 - can you maybe export the existing functions and use them? More code duplication will make merging the implementation harder.

bernardobridge:
> For this - https://github.com/artilleryio/artillery/commit/881507f826136c2dda417bc82139b1843b8f9ec3 - can you maybe export the existing functions and use them? More code duplication will make merging the implementation harder.

Yes there's a chance I can do that. I was waiting to run the full suite to see if this breaks zipfile lambdas, as then we'd have to make a change in the logic to account for it. But as we're planning on deprecating that, then maybe it doesn't matter anyway.

Will look into it after the AWS test run!

hassy:
For this https://github.com/artilleryio/artillery/commit/76df3d5d430b9c8e77834af4900ae862677238e3 - if the only difference is http-uploads, maybe it's better to change the original test and only include the plugin in a Fargate-specific environment. Less duplication would be much better.

hassy:
> But as we're planning on deprecating that, then maybe it doesn't matter anyway.

I think it matters because the two implementations will need to be merged, not just replaced. And having more of the same code in two different places will make that harder.

bernardobridge:
> I think it matters because the two implementations will need to be merged, not just replaced. And having more of the same code in two different places will make that harder.

Sorry, I wrote that in a confusing way - I was agreeing with you 👍 . I meant that we should do it, because maybe breaking zipfiles won't matter if we're going to remove them anyway.

Anyway it's done now, and it doesn't break zipfiles as long as we don't uninstall some deps we were doing before.

hassy:
LGTM!

PR #2771, opened 2024-05-21
feat(lambda): make container lambdas the default and deprecate zipfile

bernardobridge:
## Description

Deprecates `zipfile` Lambda in favour of using Lambda Container Images by default. The change should be transparent to the end-user. To make sure that's the case, the `--container` flag has been set to `deprecated` with a dedicated message (oclif functionality).

Users already using that flag will see something like this (but their test run will still run):
Screenshot 2024-05-21 at 11 38 31

## Pre-merge checklist


packages/artillery/lib/cmds/run-lambda.js
```diff
@@ -65,8 +61,11 @@ RunLambdaCommand.flags = {
     default: '1'
   }),
   container: Flags.boolean({
-    description: 'Use a container image for Lambda (experimental)',
-    default: false
+    description: '[DEPRECATED] Use a container image for Lambda (experimental)',
```
hassy, line 64:
```suggestion
    description: 'Use a container image for Lambda',
```

PR #2773, opened 2024-05-21
feat(cli): use correct architecture in worker images

bernardobridge:
## Description

We now build `x86_64` and `arm64` images. Fargate will use `x86_64` image by default (as it currently does), and Fargate will use `arm64` by default (as it currently does).

Notes:
- We don't test against `arm64` in PRs due to how long it takes to build. We will make some changes separately to run that in main against a canary
- Lambda functions will contain the architecture in the name to piggyback on existing name mechanism to determine if a new function is needed

## Pre-merge checklist

- [ ] Does this require an update to the docs? I don't think so, since no default behaviour was changed
- [ ] Does this require a changelog entry? It could be highlighted, yes

### DIFF
--- a/packages/artillery/lib/cmds/run-lambda.js
+++ b/packages/artillery/lib/cmds/run-lambda.js
@@ -66,12 +66,12 @@ RunLambdaCommand.flags = {
   }),
   container: Flags.boolean({
     description: 'Use a container image for Lambda (experimental)',
-    default: false,
-    exclusive: ['architecture']
+    default: false
   }),
   architecture: Flags.string({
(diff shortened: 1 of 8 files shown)


bernardobridge:
Note: disregard failing test, it's a bug in zipfile introduced in https://github.com/artilleryio/artillery/pull/2767/files#diff-031b22251942a138a35ea32b4c6f9f4188824a499012ab4ca3a36be352d77568L202 . 

Very easy fix (just use `noPrefix` if `noPrefixPosix` isn't there), but no point in fixing since zipfile is being deprecated in the other open PR.

PR #2778, opened 2024-05-22
fix(aws-lambda): wait a little longer for the lambda when using a VPC

bbeesley:
## Description

When using the aws-lambda platform for running tests and using a VPC, it often takes a little longer for the function to become ready that it does when creating the lambda without a VPC. When the lambda doesn't get created in time that leaves things in an inconsistent state, as the lambda is also never cleaned up. To avoid changing the behaviour when not using a VPC, this change adds a check before starting the wait for lambda flow. If the lambda is going to be connected to a VPC it sets the wait timeout to 4 minutes, otherwise it sets it to the current value of 2 minutes.

This issue happens fairly regularly, the effect being a failed test run:
image

and an orphaned lambda floating in the aws account:
image

I'm afraid I didn't open an issue first (or write tests) since the change is extremely trivial, but lmk if you think it warrants writing tests.

## Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/platform/aws-lambda/index.js
+++ b/packages/artillery/lib/platform/aws-lambda/index.js
@@ -84,6 +84,8 @@ class PlatformLambda {
       platformConfig['security-group-ids']?.split(',') || [];
     this.subnetIds = platformConfig['subnet-ids']?.split(',') || [];
 
+    this.useVPC = this.securityGroupIds.length > 0 && this.subnetIds.length > 0;
+
     this.memorySize = platformConfig['memory-size'] || 4096;
 
     this.testRunId = platformOpts.testRunId;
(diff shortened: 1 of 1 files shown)


hassy:
Looks good @bbeesley! We can merge once you've had a chance to sign the CLA. Btw we just added container Lambda support [in v2.0.12](https://github.com/artilleryio/artillery/releases/tag/artillery-2.0.12) in case you haven't seen it - should make test startup times on Lambda **much** faster.

bbeesley:
> Looks good @bbeesley! We can merge once you've had a chance to sign the CLA. Btw we just added container Lambda support [in v2.0.12](https://github.com/artilleryio/artillery/releases/tag/artillery-2.0.12) in case you haven't seen it - should make test startup times on Lambda **much** faster.

Ah nice, I'll check it out. We haven't been running as part of ci up until now to keep things fast, just been letting people run them via github actions on demand, but maybe now is a good time to revisit that.

PR #2718, opened 2024-05-09
test(publish-metrics): add tests for otel tracing

InesNi:
## Description

Implementing tests for `publish-metrics` tracing.

- A custom `FileSpanExporter` is created for testing purposes and is enabled through the `exporter` and `__outputPath` setting in the OTel reporter configuration:

```yaml copy
  plugins:
    publish-metrics:
      - type: "open-telemetry"
        traces:
          exporter: "__test"
          __outputPath: "json-file-to-export-spans-to"
```
Both the `ConsoleSpanExporter` and `InMemorySpanExporter` were considered for testing purposes but creating a custom file exporter ended up being the most straightforward option with the least possible change to the logic of the OTel reporter.

An extensive test for the `HTTP` engine tracing basic functionality support has been added including a test for the `parallel` scenario configuration and test runs with errors.

## Pre-merge checklist

- [ ] Does this require an update to the docs?
- [ ] Does this require a changelog entry?

PR #2779, opened 2024-05-22
feat(publish-metrics): implement multiple otel improvements

InesNi:
## Description
Multiple changes in OTel:
1. `url` attribute is added to page spans in Playwright tracing 

2. fix for  bug in Playwright tracing support - using `url` attribute instead of `name` is used along with `vu.uuid` to match web vitals with appropriate `pageSpan` (`replaceSpanNameRegex` introduced a bug where web vitals would not appear for spans with replaced names) 
3.  remove `test_id` being added by default to OTel metrics

4. Request timing phases  are recorded as attributes instead of spans in HTTP tracing for the purpose of reducing the cost. (e.g. `dns_lookup` span is now  `dns_lookup.duration` attribute with a value in `ms` set on the request span it belongs to.) 
    - **!This will change the structure of traces for HTTP engine support** 

### Testing

Tested manually, and will be added to OTel tracing tests in https://github.com/artilleryio/artillery/pull/2718

- Due to 4. changing the structure of tests the `dd-adot.test` that runs http test is adjusted to reflect the change in trace structure.

## Pre-merge checklist

### DIFF
--- a/packages/artillery-plugin-publish-metrics/lib/open-telemetry/metrics.js
+++ b/packages/artillery-plugin-publish-metrics/lib/open-telemetry/metrics.js
@@ -51,8 +51,7 @@ class OTelMetricsReporter {
       includeOnly: config.includeOnly || [],
       exclude: config.exclude || [],
       attributes: {
-        ...(config.attributes || {}),
-        test_id: global.artillery.testRunId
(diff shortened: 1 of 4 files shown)


hassy:
Test failure unrelated, merging.

PR #2782, opened 2024-05-23
feat: warn when a Playwright test function can't be found

hassy:
## Description

Print an error message when a Playwright test function can't be loaded. This can happen if a `testFunction` or `flowFunction` attribute specifies a function that does not exist, or if the attribute is missing.

## Pre-merge checklist

### DIFF
--- a/packages/artillery-engine-playwright/index.js
+++ b/packages/artillery-engine-playwright/index.js
@@ -352,6 +352,14 @@ class PlaywrightEngine {
           self.processor[spec.testFunction] ||
           self.processor[spec.flowFunction];
 
+        console.error('Playwright test function not found:', fn);
+        if (!fn) {
+          return cb(
+            new Error('Playwright test function not found'),
+            initialContext
+          );
+        }
+
         const test = { step };
 
         let traceScenario;


packages/artillery-engine-playwright/index.js
```diff
@@ -352,6 +352,14 @@ class PlaywrightEngine {
           self.processor[spec.testFunction] ||
           self.processor[spec.flowFunction];
 
+        console.error('Playwright test function not found:', fn);
```
HugoImaios, line 355: @hassy I might be wrong, but why is that `console.error` call not inside the if block below ?

HugoImaios: because since I updated to v2.0.13 I get spammed a **lot** about some `test function not found` :smile:

hassy: 🤦‍♂️ and this is why even "trivial" changes should have an automated test. Will release a fix asap

hassy: https://github.com/artilleryio/artillery/commit/2af901ea239a04bc42f309c4248a8ffda394aa5c

HugoImaios: no problem, it didn't break anything, just printed a lot :smile:

PR #2786, opened 2024-05-27
fix: accept 0 values for defaultNavigationTimeout and defaultTimeout

### LINKED ISSUE #2785: Artillery playwright config does not allow 0 timeout
When setting a `0` value for defaultNavigationTimeout, defaultPageTimeout or defaultTimeout, artillery fallback on 30s.

Version info: 
```
v2.0.12
```

Using this config: 
```
            defaultNavigationTimeout: 0
            defaultTimeout: 0
```

I expected to see this happen:

having a 0 timeout (no timeout)

Instead, this happened:

30s timeout is used

HugoImaios:
fixes #2785

## Description

`0` is a falsy value, so `0 || 30` will get `30` which is not what we want (`0` is a valid timeout setting, meaning no timeout).

I've used `Number` instead of `parseInt`, which is less verbose and in our case just better.
Maybe styling is not good, I'll look at the PR checks (edit: seems like they are skipped for the moment, let me know).
Speaking of lengh of lines, I found out two new questions :
- why are the setting in seconds instead of millis like the playwright config ? Isn't it just reducing the possibilities, for no gain (and les readable code here) ?
- where is `defaultPageTimeout` coming from ? It does not exist in playwright, and is not mentioned in artillery's doc neither. (so as you see I did not added a falsy protection on it)

## Pre-merge checklist

### DIFF
--- a/packages/artillery-engine-playwright/index.js
+++ b/packages/artillery-engine-playwright/index.js
@@ -14,12 +14,13 @@ class PlaywrightEngine {
     this.tracing = global.artillery.OTEL_TRACING_ENABLED || false;
 
     this.defaultNavigationTimeout =
-      (parseInt(this.config.defaultNavigationTimeout, 10) || 30) * 1000;
+      (isNaN(Number(this.config.defaultNavigationTimeout)) ? 30 : Number(this.config.defaultNavigationTimeout))
+      * 1000
+    ;
     this.defaultTimeout =
-      (parseInt(
-        this.config.defaultPageTimeout || this.config.defaultTimeout,
-        10
-      ) || 30) * 1000;
+      (isNaN(Number(
+        this.config.defaultPageTimeout || this.config.defaultTimeout
+      )) ? 30 : Number(this.config.defaultPageTimeout || this.config.defaultTimeout)) * 1000;
 
     this.testIdAttribute = this.config.testIdAttribute;
 
--- a/packages/artillery-engine-playwright/index.js
+++ b/packages/artillery-engine-playwright/index.js
@@ -18,9 +18,9 @@ class PlaywrightEngine {
       * 1000
     ;
     this.defaultTimeout =
-      (isNaN(Number(
-        this.config.defaultPageTimeout || this.config.defaultTimeout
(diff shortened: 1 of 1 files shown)


hassy:
Thanks @HugoImaios! Tests running now (they have to be manually approved to run by a core team member for external contributions for security reasons).

> why are the setting in seconds instead of millis like the playwright config

This is for consistency with elsewhere in Artillery. All other time-related config settings use seconds by default.

> where is `defaultPageTimeout` coming from ? 

This is possibly something left over from an earlier version of the plugin. I think we can remove safely it as it's not documented anywhere anyway.

HugoImaios:
>they have to be manually approved to run by a core team member

Yep, that's what I though. Thanks for launching them, even if they seem to have an issue with the artillery-cloud api key

>This is for consistency with elsewhere in Artillery. All other time-related config settings use seconds by default.

Now that the is human readable time allowed in some settings, with the ms package, I don't think that this consistency is a valid point anymore. And people using playwright engine will be used to millis, I think if it's changed [in the doc](https://www.artillery.io/docs/reference/engines/playwright#configuration) too, nobody would be surprised by inconsistency. But you have the last word :)

>This is possibly something left over from an earlier version of the plugin. I think we can remove safely it as it's not documented anywhere anyway.

agreed. Should I open a new PR after this one or remove it in this one, in another commit ?

hassy:
We still need a default unit when one is not specified, even with support for human-readable time units like `2000ms`. So for both consistency and to avoid breaking existing configs that already use seconds we'll keep it as seconds. (We really dislike breaking backwards-compatibility, which is not always easy, but then leads to nice UX like being able to pick up a third-party Artillery engine that has not been updated in 5 years, and it just works - https://evervault.com/blog/optimizing-our-encryption-engine)

For removing `defaultPageTimeout` - totally fine to do in the same PR as a separate commit!

HugoImaios:
Yes you're right, the breaking change is not worth at all :+1: 
Maybe I'll open a discussion to talk about using human-readable time units in playwright engine :slightly_smiling_face: 

>For removing defaultPageTimeout - totally fine to do in the same PR as a separate commit!
nice, will do !

hassy:
Failing tests are due to security settings on PRs from forks and not to do with the changes. Main test suite is green, merging!

PR #2800, opened 2024-05-29
ci: add GHA for publishing CF templates to S3

InesNi:
## Description
Creating a GHA for publishing 2 CloudFormation templates to S3. The templates are for creating IAM resources users need to run Lambda / Fargate tests. 

Once the templates are hosted on S3, CloudFormation quick-create links can be created and added to the docs in button format, simplifying the set up process for users.

The templates will create an an `ArtilleryDistributedTestingFargateRole`/ `ArtilleryDistributedTestingLambdaRole` with the `ArtilleryDistributedTestingFargatePolicy`/ `ArtilleryDistributedTestingLambdaPolicy` respectively.

The policies used are the original ones from the docs, though the Lambda one needed a few fixes to function properly:
- The `Resource` in the `CreateLambdaPolicy` was missing `-*` 

- The `EcrPullImagePermissions` needed to be changed so that instead of setting a `Condition` we set the `Resource` to the arn of the repository of the exact account we are pulling from

### Testing

Manually tested both Cloudformation templates making sure that all the necessary resources are created, and tested e2e by running the Fargate and Lambda tests.

Tested that the GHA appropriately updates the templates on S3.

### Notes
- Currently a test bucket is used, waiting to see if a different bucket will be used to host the templates.

## Pre-merge checklist

- [ ] Does this require a changelog entry? Not sure


packages/artillery/lib/platform/aws/iam-cf-templates/aws-iam-fargate-cf-template.json
InesNi, line 1: The policy used here is the original one from the [docs](https://www.artillery.io/docs/load-testing-at-scale/aws-fargate#iam-permissions), the only difference is the account id being set dynamically.

The role is currently set to be able to be assumed by anyone/anything and should be appropriately restricted by user.

packages/artillery/lib/platform/aws/iam-cf-templates/aws-iam-lambda-cf-template.json
```diff
@@ -0,0 +1,108 @@
...
+                    "iam:PassRole",
+                    "iam:AttachRolePolicy"
+                  ],
+                  "Resource": {"Fn::Sub": "arn:aws:iam::${AWS::AccountId}:role/artilleryio-default-lambda-role-*"}
+                },
+                {
+                  "Sid": "CreateLambdaPolicy",
+                  "Effect": "Allow",
+                  "Action": ["iam:CreatePolicy"],
+                  "Resource": {"Fn::Sub": "arn:aws:iam::${AWS::AccountId}:policy/artilleryio-lambda-policy-*"}
```
InesNi, line 44: In the original policy from the docs `Resource` here was missing `-*` at the end.

packages/artillery/lib/platform/aws/iam-cf-templates/aws-iam-lambda-cf-template.json
```diff
@@ -0,0 +1,108 @@
...
+                  "Resource": {"Fn::Sub": "arn:aws:lambda:*:${AWS::AccountId}:function:artilleryio-*"}
+                },
+                {
+                  "Sid": "EcrPullImagePermissions",
+                  "Effect": "Allow",
+                  "Action": [
+                    "ecr:GetDownloadUrlForLayer",
+                    "ecr:BatchGetImage"
+                  ],
+                  "Resource": "arn:aws:ecr:*:248481025674:repository/artillery-worker"
```
InesNi, line 76: The `EcrPullImagePermissions` needed to be changed so that instead of setting a `Condition` we set the `Resource` to the arn of the repository of the exact account we are pulling from.

The initial policy from the docs was using:
```
{
  "Sid": "EcrPullImagePermissions",
  "Effect": "Allow",
  "Action": [
                    "ecr:GetDownloadUrlForLayer",
                    "ecr:BatchGetImage"
                  ],
  "Resource": "*",
  "Condition": {
     "StringLike": {
          "aws:sourceArn": "arn:aws:lambda:*:123456789000:function:artilleryio-*"
        }
  }
}
```

.github/workflows/s3-publish-cf-templates.yml
```diff
@@ -1,15 +1,34 @@
 name: Publish CloudFormation templates to AWS S3
 
 on:
+  push:
+    branches:
+      - main
+    paths:
+      - './packages/artillery/lib/platform/aws/iam-cf-templates/**'
```
bernardobridge, line 8: We'll have to change this, as the CF stack that is in main will be for an unreleased version of the CLI.

We either do `workflow_dispatch` only, but then we need to be able to specify which artillery version to test against (e.g. the latest canary), or we do `workflow_call` and trigger this for canary and for mainline.

In either case, it sounds like we might need to actually publish two versions of the templates, one for the canary, and one for the latest.

bernardobridge (changes requested):

bernardobridge (approved):

PR #3190, opened 2024-06-10
fix: handle SIGTERM correctly on Fargate

hassy:
## Description

Fix handling of `SIGTERM`:

- Exit with appropriate exit code (143) from the `run` command
- On Fargate (e.g. when stopped early by Fargate Spot, or via explicit task stop, e.g. through `aws ecs stop-task`) - replace the use of `wait` with `timeout`. `wait` cannot be used for processes that run in a subshell - https://github.com/artilleryio/artillery/blob/27fae398ce0d6bab5f4226e361a9de6d4d8bec2a/packages/artillery/lib/platform/aws-ecs/worker/loadgen-worker#L277

## Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/cmds/run.js
+++ b/packages/artillery/lib/cmds/run.js
@@ -337,10 +337,10 @@ RunCommand.runCommandImplementation = async function (flags, argv, args) {
     var finalReport = {};
     var shuttingDown = false;
     process.on('SIGINT', async () => {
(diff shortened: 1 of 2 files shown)


hassy:
Existing E2E tests passing on Fargate. Adding an E2E test for the error condition needs an easy way to get the task ARNs for a running test, so leaving for later. Tested by stopping a task in a running test manually and everything looks good.

PR #3219, opened 2024-06-20
feat: a few ECS-related tweaks

hassy:
## Description

A couple of ECS related things:

* Follow up to #3192. The fix in that PR did not work as intended because the array would always contain 2 elements
* Introduce `run-ecs` alias for `run-fargate`. This will be useful in future to make things less ambiguous when we add support for Amazon EKS, which also supports Fargate as a capacity provider.
* A couple of tweaks/fixes to copy

## Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
+++ b/packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
@@ -1170,34 +1170,37 @@ async function ensureTaskExists(context) {
         };
       });
 
+    const artilleryContainerDefinition = {
+      name: 'artillery',
+      image: imageUrl,
+      cpu: cpu,
+      command: [],
+      entryPoint: ['/artillery/loadgen-worker'],
+      memory: memory,
+      secrets: secrets,
+      ulimits: ulimits,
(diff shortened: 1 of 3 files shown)


packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
```diff
@@ -1170,35 +1170,38 @@ async function ensureTaskExists(context) {
...
-            options: {
-              'awslogs-group': `${context.logGroupName}/${context.clusterName}`,
-              'awslogs-region': context.region,
-              'awslogs-stream-prefix': `artilleryio/${context.testId}`,
-              'awslogs-create-group': 'true',
-              mode: 'non-blocking'
-            }
-          }
-        },
-        ...([context.adot?.taskDefinition] || [])
```
hassy, line 1197: This would always add `undefined` to the resulting array, which means that the check on [line 1204](https://github.com/artilleryio/artillery/pull/3219/files#diff-6ffb0c78caac19b3553805ce01e10b97aea363c2862007d099d09100d91c0150L1204) would always evaluate to `true`.

PR #3232, opened 2024-06-26
fix(http): process file uploads correctly

InesNi:
## Description

As the `http-file-uploads` plugin is outdated, file uploads will now be directly supported in the HTTP engine instead.  

### Configuration
Files can be uploaded as part of the `formData` option by adding the `fromFile: ./path/to/your/file` as a field value to the files field name key as seen in example below
```yaml copy
scenarios:
  - name: "account"
    flow:
    - post:
        url: "/upload"
        formData:
          id: 12345
          name: "George"
          avatar:    # file fields key
            fromFile: "./path/to/avatar.jpg" # file fields value
   
```
When running tests in distributed mode on Fargate/Lambda, the files will need to be added to `includeFiles` config option as well.
```yaml copy
config:
  target: "http://localhost:3007"
  phases:
    - duration: 1
      arrivalRate: 1
  includeFiles:
    - "./path/to/avatar.jpg"
    
scenarios:
  - name: "account"
    flow:
    - post:
        url: "/upload"
        formData:
          id: 12345
          name: "George"
          avatar:
            fromFile: "./path/to/avatar.jpg"
   
```
Additionally, the `setContentLengthHeader` option from plugin will also be available as a request level configuration option and will only be functional for requests with file uploads. (This is here for S3/CloudWatch pre-sign URLs and assumes that only one file is being uploaded)
```yaml copy
scenarios:
  - name: "account"
    flow:
    - post:
        url: "/upload"
        setContentLengthHeader: true
        formData:
          id: 12345
          name: "George"
          avatar:
            fromFile: "./path/to/avatar.jpg"
   
```
### Implementation
**File path resolution**
Since the core package is not aware of the location of the file that holds `scenarios`, a `$scenarioFile` variable has been added to [test level variables](https://www.artillery.io/docs/reference/test-script#test-level-variables).  To achieve that:
- an additional `scriptPath` parameter has been added to `resolveConfigTemplates` function where it is added to the `context.vars` as `$scenarioFile`
- the path to `scenarios` has been passed as an additional argument to the `resolveConfigTemplates` function implementations (from Lambda, Fargate and local)
- `$scenarioFile` has been added to `INITIAL_CONTEXT` in `runner.js` to make it available to `scenarios` part of script

**File upload logic**
- The file uploads are handled directly in the `engine_http.js`:
    -  We check for `fromFile` variable in `formData`,
    -  When `fromFile` is present its value is resolved to the absolute path 
    -  A readable stream of the file is created and replaces the `{fromFile: path/to/file}` as a field value. 
    -  The field (`fieldName: fieldValue`) proceeds to be appended to the `form-data` object along with other fields.
    -  If `setContentLengthHeader` is set to `true` we get the size of the file and set the header

**Additional changes**
- Example for `file-uploads` has been changed accordingly and renamed to `http-file-uploads` to match the pattern in examples
- The schema has been updated for `includeFiles` and `setContentLengthHeader`,  however the `formData` will be adjusted in a separate PR

### Testing

- The file upload is inherently a core HTTP engine feature, however it does rely on certain elements made available by integrating with artillery package (such as `$scenarioFile` variable). Due to this to make the test e2e, it was placed in the artillery package
- The test was placed in the `integration/core` folder as it did not really belong in existing folders, and the `integration ` folder can now be used to sort other tests (e.g. `publish-metrics`)

#### Checks
- Checks that:
    - The request was successful
    - Both file and non-file fields have been sent
    - Each field has appropriate values
    - The files were sent correctly (creates a hash of each file and checks against the hash of files received by the server)

#### Components
- Test file - spins server, runs artillery test and performs assertions
- Server
    -  A file upload server that can take multiple file fields as well as other fields.
    - Hashes received files so we can compare it against the hash of the files we sent.
    - Responds with the object containing status, files and other fields
    -  uses `hapi` library that already exists as dependency
- YAML script - configured to send HTTP POST request a form with multiple file fields and a non-file field
- Processor - Logs server response body to console so we can parse it from the test
- Files to upload - a jpg and pdf file

## Pre-merge checklist

### To be added to docs
- `fromFile`
- `setContentLengthHeader`
- `$scenarioFile`
- add files to `includeFiles` when running on AWS


packages/core/lib/engine_http.js
```diff
@@ -399,12 +401,22 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
           requestParams.body = _.reduce(
             requestParams.formData,
             function (acc, v, k) {
-              // acc[k] = template(v, context);
-              acc.append(k, template(v, context));
+              let V = template(v, context);
+              if (V && _.isPlainObject(V) && V.fromFile) {
+                V = fs.createReadStream(
+                  path.resolve(process.cwd(), V.fromFile)
```
hassy, line 407: This needs to be resolved relative to the test script itself (same way we're doing it for `processor` or `payload` entries for example). Otherwise the test script will work only when the CLI is run from a specific directory.

packages/core/lib/engine_http.js
```diff
@@ -399,12 +401,22 @@ HttpEngine.prototype.step = function step(requestSpec, ee, opts) {
...
+              }
+              acc.append(k, V);
               return acc;
             },
             f
           );
+          if (params.setContentLengthHeader) {
+            requestParams.headers = requestParams.headers || {};
+            requestParams.headers['content-length'] =
+              requestParams.body.getLengthSync();
```
hassy, line 418: This doesn't work - https://github.com/form-data/form-data?tab=readme-ov-file#integer-getlengthsync

hassy:
Multiple files don't seem to work, e.g. with:

```yaml
    - post:
        url: "/upload"
        formData:
          document:
            fromFile: "./files/sre-fundamental-rules.png"
          anotherDocument:
            fromFile: "./files/artillery-logo.jpg"
```

packages/artillery/lib/util.js
```diff
@@ -98,11 +98,12 @@ function addDefaultPlugins(script) {
   return finalScript;
 }
 
-async function resolveConfigTemplates(script, flags, configPath) {
+async function resolveConfigTemplates(script, flags, configPath, scriptPath) {
```
InesNi, line 101: We add the additional argument here for the path to script (file that holds `scenarios`) and set it to `vars` making it available at config level

packages/artillery/lib/util.js
```diff
@@ -198,7 +199,7 @@ async function checkConfig(script, scriptPath, flags) {
     );
     payloadSpec.path = resolvedPathToPayload;
   });
-
+  script._scriptPath = absoluteScriptPath;
```
InesNi, line 202: Adding the `scriptPath` to the `script` in `checkConfig` function so we can feed it to the `resolveConfigTemplates` from `run.js` to make it available for local runs.

Follows the same process that was done for existing `$dirname` variable.

packages/artillery/lib/cmds/run.js
```diff
@@ -498,10 +498,12 @@ async function prepareTestExecutionPlan(inputFiles, flags, args) {
 
   const script3 = await addOverrides(script2, flags);
   const script4 = await addVariables(script3, flags);
+  // The resolveConfigTemplates function expects the config and script path to be passed explicitly because it is used in Fargate as well where the two arguments will not be available on the script
   const script5 = await resolveConfigTemplates(
     script4,
     flags,
-    script4._configPath
+    script4._configPath,
+    script4._scriptPath
```
InesNi, line 506: Passing the path to `scenarios` here making it available in config for local runs

packages/artillery/lib/platform/aws-ecs/legacy/bom.js
```diff
@@ -34,7 +35,8 @@ function createBOM(absoluteScriptPath, extraFiles, opts, callback) {
           opts: {
             scriptData,
             absoluteScriptPath,
-            flags: opts.flags
+            flags: opts.flags,
+            scenarioPath: opts.scenarioPath // Absolute path to the file that holds scenarios
```
InesNi, line 39: Adding the `scenarioPath` to options here so we can access and pass it to `resolveConfigTemplates` making it available in config for distributed runs.

Both Lambda and Fargate use this function and will pass the path to scenarios to `opts` from their side

packages/core/lib/engine_http.js
InesNi: Addressed this by making path to scenarios available as test level variable `$scenarioFile` as discussed. Commit https://github.com/artilleryio/artillery/pull/3232/commits/4dd0f57a4af69e5be938320e647688cd3974b17c

packages/core/lib/engine_http.js
InesNi: Resolved using `fs` like plugin does. Commit https://github.com/artilleryio/artillery/pull/3232/commits/b7e07115bbf34c9513ac21bc88101c6603f2aca9

InesNi:
> Multiple files don't seem to work, e.g. with:
> 
> ```yaml
>     - post:
>         url: "/upload"
>         formData:
>           document:
>             fromFile: "./files/sre-fundamental-rules.png"
>           anotherDocument:
>             fromFile: "./files/artillery-logo.jpg"
> ```

Multiple files were working but the e2e test server was testing for single file. A new test has been added with a different server, covering functionality of multiple file uploads.

PR #3270, opened 2024-07-16
fix(fargate): set cloud endpoint from env in fargate containers

bernardobridge:
## Description

This fixes two issues:
- Cloud workers will do the `/whoami` call in the container without taking into account the endpoint, so if the endpoint is not prod, the api key will not be correct, causing the error.
- If you set `ARTILLERY_CLOUD_API_KEY` in the environment, but don't set `--record`, the key will still get passed to the worker, which causes it to run unnecessarily.

## Pre-merge checklist


InesNi (commented):
Lgtm!

InesNi (approved):

Issue #3268, opened 2024-07-13
Bug:  Payloads with `loadAll` and field names include unexpected objects


Rio517:
Hello!

When using `loadAll` with field names, `vuContext.vars` includes the expected array of objects and each field name loading with an object from the CSV rows.

Version info:
```
Artillery: 2.0.17
Node.js:   v20.15.0
OS:        linux
Playwright Version 1.45.1
```

Running this command:
```
npx artillery run my-test.yml
```

## Expected to see this happen:

In `vuContext.vars`
```js
{
  payloadName: [
    { field1: '01', field2: '01' },
    { field1: '02', field2: '02' },
    { field1: '03', field2: '03' },
    ...
  ]
}
```
## Instead, this happened:

In `vuContext.vars`

```js
{
  field1: { field1: '01', field2: '01' },
  field2: { field1: '02', field2: '02' },
  payloadName: [
    { field1: '01', field2: '01' },
    { field1: '02', field2: '02' },
    { field1: '03', field2: '03' },
    ...
  ]
} 
```

**Note:** This becomes super problematic when using multiple identical field names from different payloads. I had two payloads, one normal payload and one with `loadAll`. Both payloads had an `account_id`.  In theory, only the non-`loadAll` payload's `account_id` should have been at the root of `vuContext.vars`. However, the unintended `account_id` from the `loadAll` payload would overright the first value - leading to confusion.

Theoretically, one could reorder the payloads, but I wasn't sure how deterministic that was. I just renamed the keys once I figured out what was going on. 

### Simplified configuration

````yaml
config:
  engines:
    playwright: {}
  payload:
    path: 'path/contacts.csv'
    fields:
      - 'first_name'
      - 'last_name'
    loadAll: true
    name: 'uploadContacts'
scenarios:
- engine: playwright
  name: 'ContactUpload'
  testFunction: 'contactUploadTest'
````

### console.log(vuContext.vars);
````js
{
  first_name: {
    first_name: 'load_first_01',
    last_name: 'load_last_01',
  },
  last_name: {
    first_name: 'load_first_02',
    last_name: 'load_last_02',
  }
  uploadContacts: [
    {
      first_name: 'load_first_01',
      last_name: 'load_last_01',
    },
    {
      first_name: 'load_first_02',
      last_name: 'load_last_02',
    },
    {
      first_name: 'load_first_03',
      last_name: 'load_last_03',
    },
    {
      first_name: 'load_first_04',
      last_name: 'load_last_04',
[...]

PR #3277, opened 2024-07-22
fix(core): prevent setting random fields when using loadAll

bernardobridge:
## Description

- Should solve https://github.com/artilleryio/artillery/issues/3268
- Adds a warning when `loadAll` is set without `name`

## Testing
Automated tests need to be added for all the possible options (there aren't any existing ones that check these options). I'll handle that separately to get the fix out now.

However, I have tested manually all cases:
- `loadAll:true, name:something`: this is the bug, resolves it by not setting the random key:value pairs on the root, and only setting the whole csv
- `loadAll:undefined, name:something`: remains the same (sets the data under the name key, but also sets the random key:value pairs).
- `loadAll:true, name:undefined`: previously would set a random key:value pair on the root. This doesn't make sense, as there is a clear intent to load all data. Now logs a warning (in the future, perhaps this should be done with earlier validation, but no validation exists currently for any payload fields).

I've also tested the above when no `fields` are set, and it works (with `loadAll` + `name` it loads all the data, with just `name` it loads a random row).

## Pre-merge checklist

- [ ] Does this require an update to the docs? Yes, this section should be clarified

### DIFF
--- a/packages/core/lib/runner.js
+++ b/packages/core/lib/runner.js
@@ -405,18 +405,27 @@ function datafileVariables(script) {
   let result = {};
   if (script.config.payload) {
     _.each(script.config.payload, function (el) {
-      // If data = [] (i.e. the CSV file is empty, or only has headers and
-      // skipHeaders = true), then row could = undefined
-      let row = el.reader(el.data) || [];
-      _.each(el.fields, function (fieldName, j) {
-        result[fieldName] = row[j];
-      });
+      //when loading all the csv, we don't set individual fields
+      if (!el.loadAll) {
+        // If data = [] (i.e. the CSV file is empty, or only has headers and
+        // skipHeaders = true), then row could = undefined
+        let row = el.reader(el.data) || [];
+        _.each(el.fields, function (fieldName, j) {
+          result[fieldName] = row[j];
+        });
+      }
+
       if (typeof el.name !== 'undefined') {
         // Make the entire CSV available
         result[el.name] = el.reader(el.data);
+      } else {
+        console.log(
+          'WARNING: loadAll is set to true but no name is provided for the CSV data'
+        );
       }
     });
   }
+
   return result;
 }
 


hassy:
The issue in #3268 is that variables created from field names (`field1`, `field2`) are set to the value of the entire row, rather than to individual columns. This change seems to ignore the `fields` attribute if `loadAll` is set, rather than assigning those variables the correct values.

To use one of our official examples: https://github.com/artilleryio/artillery/blob/main/examples/table-driven-functional-tests/functional-test.yml#L5

After this change `url` and `code` scenario variables won't be set at all, only a `data` variable.

bernardobridge:
> The issue in #3268 is that variables created from field names (`field1`, `field2`) are set to the value of the entire row, rather than to individual columns. This change seems to ignore the `fields` attribute if `loadAll` is set, rather than assigning those variables the correct values.

Right @hassy, but that's only half the issue. Even if we assign those variables the "correct" values, if the user is using more than one payload, and those payloads have similar fields, then they can override each other. To quote the user:

>Both payloads had an account_id. In theory, only the non-loadAll payload's account_id should have been at the root of vuContext.vars. However, the unintended account_id from the loadAll payload would overright the first value - leading to confusion.

So it seemed better to go by the assumption that when the user has the intent to use `loadAll`, they are looking to have everything from the CSV available, not a random value from the data.

---

> To use one of our official examples: https://github.com/artilleryio/artillery/blob/main/examples/table-driven-functional-tests/functional-test.yml#L5
> 
> After this change `url` and `code` scenario variables won't be set at all, only a `data` variable.

Yes, but the example will still work correctly (as it loops over the `data` property.)

hassy:
So we need to decide what happens if multiple payloads are defined with the same field names:

1. Keep Artillery's current behavior which always override the variables, or:
2. Change it so that conflicting field names still override each other **unless** a payload definition has `loadAll: true` set. In which case the `fields` spec on that payload will be ignored.

To me (1) is easier to reason about. We just need to fix the bug where field variables are set to an entire row rather than a single value. The changes in the PR only address a special case of clashing field/variable names.

> Yes, but the example will still work correctly (as it loops over the data property.)

Yes the test itself doesn't use those variables, but I linked to it as an example of the behavior that will change with this PR.

bernardobridge:
We discussed this internally, and decided to keep the current PR as is, because `loadAll` declares an intent to use the entire CSV, and the user can then choose rows from it as they decide.

Guzikowski:
@bernardobridge I have not used loadall, but now getting this warning that is set to true than a CSV must b defined. Is there a way to not show this warning? I am assuming if I set to the loadall to false then I would not see the error, but no luck.

hassy:
thanks @Guzikowski, can confirm it's a regression. We'll release a fix shortly

PR #3293, opened 2024-07-29
fix(core): import mjs path correctly to account for windows paths

bernardobridge:
## Description

Same fix as https://github.com/artilleryio/artillery/pull/2662. PR was inactive , and we've since implemented windows tests in CI to confirm the fix, so easier to just create the PR ourselves to confirm the fix.

## Pre-merge checklist


bernardobridge:
Closed in favour of the original PR (https://github.com/artilleryio/artillery/pull/2662), as the author has now rebased.

PR #3291, opened 2024-07-26
feat: Azure support

hassy:
## Description

Add support for running Artillery tests on Azure ACI.

## Pre-merge checklist

**This is for use by the Artillery team. Please leave this in if you're contributing to Artillery.**

- [ ] Does this require an update to the docs? Yes
- [ ] Does this require a changelog entry? Yes

PR #3300, opened 2024-08-06
fix(lambda): set plugin path to plugin folder and test data folder

bernardobridge:
## Description

https://github.com/artilleryio/artillery/pull/3291 introduced a regression to Lambda tests by [removing the installation](https://github.com/artilleryio/artillery/pull/3291/files#diff-a33800f8944e9e0680e8427bdbe98ef51764c448e6681c5a68737ca2c304fc19L43) of the `artillery-plugin-sqs-reporter`. This was anyway a piece of tech debt, as it's better to set the path to the plugin code that is bundled with the Artillery source code.

## Pre-merge checklist

- [ ] Does this require a changelog entry? No, this was never released


hassy:
🙌

PR #3301, opened 2024-08-06
feat(ecs): add ability to set DNS servers and ephemeral storage size

hassy:
## Description

Add ability to customize two settings for Artillery workers on ECS:

- `--container-dns-servers` can now be used to set the `dnsServers` setting for the Artillery container
- `--ephemeral-storage` can now be used to set the amount of ephemeral storage for the worker task

## Pre-merge checklist


hassy:
Tested manually:
- ephemeral storage only, on Fargate (only supported on Fargate)
- DNS servers only, on ECS/EC2 (not supported when network mode = `awsvpc` which is what we set on Fargate)

PR #3311, opened 2024-08-13
feat(JSONPath-Plus): Use library correctly

Archangelza1:
## Description

This updates the JSON parse logic so that it returns what is requested by the user using the [JSONpath-plus](https://www.npmjs.com/package/jsonpath-plus) standard queries

refactors the `extractJSONPath` function to improve its flexibility and simplify its implementation.

## Pre-merge checklist

### DIFF
--- a/packages/commons/engine_util.js
+++ b/packages/commons/engine_util.js
@@ -271,8 +271,7 @@ function templateObjectOrArray(o, context) {
     debug(
       `path = ${path} ; value = ${JSON.stringify(
         value
-      )} (${typeof value}) ; (subj type: ${
-        subj.length ? 'list' : 'hash'
+      )} (${typeof value}) ; (subj type: ${subj.length ? 'list' : 'hash'
       }) ; newValue = ${JSON.stringify(newValue)} ; newPath = ${newPath}`
     );
 
@@ -571,7 +570,6 @@ function dummyParser(body, callback) {
   return callback(null, body);
 }
 
-// doc is a JSON object
 function extractJSONPath(doc, expr) {
   // typeof null is 'object' hence the explicit check here
   if (typeof doc !== 'object' || doc === null) {
@@ -581,7 +579,7 @@ function extractJSONPath(doc, expr) {
   let results;
 
   try {
-    results = jsonpath(expr, doc);
+    results = jsonpath({ path: expr, json: doc, wrap: false });
   } catch (queryErr) {
     debug(queryErr);
   }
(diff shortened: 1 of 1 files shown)


packages/commons/engine_util.js
```diff
@@ -590,11 +588,7 @@ function extractJSONPath(doc, expr) {
     return '';
   }
 
-  if (results.length > 1) {
-    return results[randomInt(0, results.length - 1)];
-  } else {
-    return results[0];
-  }
+  return results;
```
hassy, line 591: This is a breaking change to current behavior (we try really hard to not break backwards-compatibility). Perhaps the behavior can be configurable via a parameter to the `json` attribute? e.g.

```yaml
capture:
  - json: '$.results[*]'
    as: results
    multiple: true # return all matches
```

packages/commons/engine_util.js
```diff
@@ -581,7 +579,7 @@ function extractJSONPath(doc, expr) {
   let results;
 
   try {
-    results = jsonpath(expr, doc);
+    results = jsonpath({ path: expr, json: doc, wrap: false });
```
hassy, line 582: Setting `wrap` to `false` is also a breaking change. Maybe that could be configurable as an option too.

hassy:
Thanks for the PR @Archangelza1! We can't merge it as-is as it would break backwards-compatibility, but the new behavior could be introduced via extra configuration.

packages/commons/engine_util.js
Archangelza1: I have updated this to be backwards compatible

packages/commons/engine_util.js
Archangelza1: Please see the changes I have made

Archangelza1:
@hassy I have made an update to the code to allow backwards compatibility

Archangelza1:
@hassy Any update on this PR?

hassy:
hi @Archangelza1 apologies for the delay on this one. Thanks for making the changes but it doesn't get us there completely:

1. Current behavior is maintained when `multiple` is not set, that's good.
2. Setting `multiple` to `true` is now equivalent to current behavior
3. Setting `multiple` to `false` will return all results - this is not very intuitive

Ideally not setting `multiple` at all should be equivalent to setting `multiple` = false, and setting `multiple` to true should add new behavior.

I'll merge this and fix it up in a separate commit.

hassy:
Failing tests are ones that rely on a secret which is not passed to CI runs from forks. The core test suite is passing.

Issue #3203, opened 2024-06-17
Error ERR_REQUIRE_ESM on attempt to use a ES module with *.js extension as a processor


drdaemos:
Version info: 
```
Artillery: 2.0.14
Node.js:   v20.9.0
OS:        linux
```

Running this command: 
```
artillery run test.yml
```

I expected test to run successfully, but
Instead, this happened:

```
⠋ /<...>/node_modules/@artilleryio/int-core/lib/runner.js:97
      script.config.processor = require(processorPath);
                                ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /<...>/processor.js from /<...>/node_modules/@artilleryio/int-core/lib/runner.js not supported.
Instead change the require of processor.js in /<...>/node_modules/@artilleryio/int-core/lib/runner.js to a dynamic import() which is available in all CommonJS modules.
    at loadProcessor (/<...>/node_modules/@artilleryio/int-core/lib/runner.js:97:33)
```

Files being used:

[test.yml]
```
config:
  # ...
  processor: 'processor.js' # pointed to a ES Module
```

My project has ES modules with `*.js` extension and `type="module"` set in the `package.json`. According to https://nodejs.org/api/packages.html#determining-module-system it is a valid way of specifying ES module files.

Problem arises from the following code in `packages/core/lib/runner.js`:

```
async function loadProcessor(script, options) {
  const absoluteScriptPath = path.resolve(process.cwd(), options.scriptPath);
  if (script.config.processor) {
    const processorPath = path.resolve(
      path.dirname(absoluteScriptPath),
      script.config.processor
    );

    if (processorPath.endsWith('.mjs')) { // <- this check is causing to divert to incorrect path
      const exports = await import(processorPath);
      script.config.processor = Object.assign(
        {},
        script.config.processor,
        exports
      );
    } else {
      // CJS (possibly transplied from TS)
      script.config.processor = require(processorPath);
    }
  }

  return script;
}
```

Probable fix could be simply removing the conditional and always attempt to `import(module)` as Node.js should be able to determine the module format and import it correctly for both CJS and ESM.

PR #3405, opened 2024-11-13
fix: add utility functions to get  module type

jonasclaes:
## Description

This PR will add more complete handling with regards to the CommonJS and ESM standards. This PR should solve #3203 and #2661.

We now look at the extension first, if it's .cjs or .mjs, great, that's easy to resolve. Otherwise, we look for the first package.json we come across, check it's module type, and then determine the project's standards.

The esbuild has been updated to accomodate this change as well.

## Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/util/prepare-test-execution-plan.js
+++ b/packages/artillery/lib/util/prepare-test-execution-plan.js
@@ -2,6 +2,7 @@ const csv = require('csv-parse');
 const fs = require('node:fs');
 const path = require('node:path');
 const p = require('util').promisify;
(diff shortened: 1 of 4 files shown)


hassy:
thank you @jonasclaes! will aim to review & release in the next version of Artillery

jonasclaes:
Hey @hassy!

It's been almost two months since the opening of this PR now. Do you have an estimate of when this next version will be released including this change?

Thanks in advance!

jonasclaes:
Closed due to non-response.

PR #3445, opened 2025-01-09
Pre-install browsers in docker container

andrewvc:
The current docker container can't run playwright scripts since the required playwright browsers are missing, this PR fixes that by pre-installing them

## Pre-merge checklist

### DIFF
--- a/packages/artillery/Dockerfile
+++ b/packages/artillery/Dockerfile
@@ -1,11 +1,11 @@
-FROM node:18-alpine
+FROM node:22-bookworm-slim
 LABEL maintainer="team@artillery.io"
 
 WORKDIR /home/node/artillery
 
 COPY package*.json ./
 RUN npm --ignore-scripts --production install
-RUN yes | npx playwright install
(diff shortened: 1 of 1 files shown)


hassy:
Thanks @andrewvc!

hassy:
We'll need to make a couple of changes to the Dockerfile before shipping this:

- Current Docker image is the plain Alpine-based Node.js image. It won't have the OS dependencies needed to run the browsers, so we'll need to install those. `playwright install --with-deps` will do it for us, but it expects an Ubuntu/Debian-based system, so:
  - We'll need to switch to using an Ubuntu/Debian-based Node.js image. [`22-bookworm-slim`](https://hub.docker.com/_/node) should work.
- Artillery's Playwright integration is [limited to Chromium](https://www.artillery.io/docs/playwright#limits--gotchas) at the moment, so we don't need to install all default browsers. `playwright install chromium` will install just Chromium.

andrewvc:
Thank you!

PR #3499, opened 2025-03-18
feat: Slack plugin improvements

manfromanotherland:
## Description

When a large amount of checks are defined, users were hitting a Slack error around the message length limit. The same was also happening when a test would have a large amount of errors.

- Adds limits around showing checks and errors, show summary if too many
- Improve message formatting following Slack guidelines

## Preview

| Before | After |
| --- | --- |
| before | after |
| before | after |
| before | after |

## Pre-merge checklist


hassy:
🚀

PR #3651, opened 2025-11-10
fix: pass --scenario-name through on Fargate

hassy:
## Description

`--scenario-name` is not currently passed through correctly into Fargate workers, because CLI flags are converted to camelCase in: https://github.com/artilleryio/artillery/blob/aa9e68771afd4666da58d6f3938e861869f72d98/packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js#L136-L140.

## Pre-merge checklist

### DIFF
--- a/packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
+++ b/packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
@@ -1426,7 +1426,7 @@ async function generateTaskOverrides(context) {
     context.cliOptions.environment
       ? ['--environment', context.cliOptions.environment]
       : [],
-    context.cliOptions['scenario-name']
+    context.cliOptions['scenarioName']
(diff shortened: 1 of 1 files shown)


packages/artillery/lib/platform/aws-ecs/legacy/run-cluster.js
```diff
@@ -1426,7 +1426,7 @@ async function generateTaskOverrides(context) {
     context.cliOptions.environment
       ? ['--environment', context.cliOptions.environment]
       : [],
-    context.cliOptions['scenario-name']
+    context.cliOptions['scenarioName']
       ? ['--scenario-name', context.cliOptions['scenario-name']]
```
barth12: @hassy shouldn't it be `? ['--scenario-name', context.cliOptions['scenarioName']]`. I checked the fix and unfortunately it does not work

PR #3653, opened 2025-11-12
fix(bom): resolve payload paths in separate config

hassy:
## Description

Ensure that payload paths are resolved relative to the main script to match behavior when executing the test locally.

## Pre-merge checklist


hassy:
E2E test failure unrelated, will open separate PR to fix

